PR #5, opened 2023-04-08
pch + by

vincentarelbundock:
This PR implements the @zeileis suggestion to make `pch` a bit smarter when using the `by` argument. See <https://github.com/grantmcdermott/plot2/issues/4#issuecomment-1500741433>

I could easily do something similar for `col` if you think this is something you would eventually want to merge.

``` r
library(plot2)

# default
with(mtcars, plot2(mpg ~ hp | gear))
```

``` r

# manual shape: vector must be same length as number of groups
with(mtcars, plot2(mpg ~ hp | gear, pch = 2:4))
```

``` r

# user input sanity check
with(mtcars, plot2(mpg ~ hp | gear, pch = 1:32))
# Error: `pch` must be `NULL` or a vector of length 1 or 3.
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -5,10 +5,18 @@ Version: 0.0.1
 Date: 2023-02-21
 Authors@R: 
   c(
-    person(given = "Grant",
-           family = "McDermott",
-           role = c("aut", "cre"),
-           email = "grantmcd@uoregon.edu"
+    person(
+      given = "Grant",
+      family = "McDermott",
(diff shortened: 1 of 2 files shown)


grantmcdermott:
Dig it. Thanks @vincentarelbundock, please add your name as a contributor in the DESCRIPTION and I'll look to merge.

(As an aside, I still need to add the tinysnapshot testing scaffolding but will try to do that when I get a sec today.)

PR #8, opened 2023-04-08
plot2.formula now does proper formula parsing via stats::model.frame

zeileis:
To address issue https://github.com/grantmcdermott/plot2/issues/3, I have implemented now a proper formula parsing for `plot2.formula`:

* Internally `y ~ x | z` is simply converted to `y ~ x + z`.
* Then `stats::model.frame()` is called "as usual" with `subset`, `na.action`, and `drop.unused.levels`.
* The first variable from the resulting model frame is used as `y`, the second as `x`, and all others (if any) are collapsed to a single factor. Either via `as.factor()` (for a single variable) or `interaction()` (if more than one variable).
* Default axis and legend labels are taken from the variable names of the model frame.

Open question: Many arguments of `plot2.formula()` are actually not used at all in `plot2.formula()` but only passed on to `plot2.default()`. If these arguments have defaults that are identical between the two methods, the argument could also be omitted from `plot2.formula()`. If needed it is passed through `...`.


grantmcdermott:
Excellent, thanks @zeileis!

Do you mind quickly adding yourself as contributor in the DESCRIPTION file? We can merge directly after.

RE: "Unused" arguments in `plot2.formula`. You're quite right. I think that's just a vestige of my original script &mdash;before I packaged it up&mdash;where the formula implementation was the sole default without any method dispatch. We can change it as part of this PR or separately. Up to you.

grantmcdermott:
Had a chance to update the DESCRIPTION file. Let's take the argument passing separately.

PR #9, opened 2023-04-09
Drop unneeded args

grantmcdermott:
@zeileis Here's a version that removes the duplicate arguments in `plot2.formula`  to pass everything through `...`. One thing thing makes me hesitate is that this masks argument autocompletion. Specifically, the dropped arguments like `legend.position` don't appear/autocomplete once when users start calling `plot2(<some formula>, ...)`. 

Note that this might be limited to RStudio. Some of the other editors/interfaces that I tried (e.g. Vim and radian) don't seem be affected by this masking. But given that RStudio is the modal editor for R users, I'm a little hesitant to merge this PR.

Update: And checks are failing. I'll look at this again tomorrow.


zeileis:
Autocompletion is a a good point, especially given that convenience and easy of use for beginners is fundamental to the package.

grantmcdermott:
Tabling this for now.

Issue #7, opened 2023-04-08
Should default legend position be "right!" instead of "bottom!"?


grantmcdermott:
Would make it consistent with **ggplot2** and also seems more logical given the landscape orientation / aspect ratio of most plots. I'm leaning that way, but happy for others to way in with thoughts.

zeileis:
I agree, my preference for the default position is also "right".

zeileis:
Another question: Personally, I loathe the boxes around the base legends. In my opinion, `bty = "n"` would be the better default. What do you think?

grantmcdermott:
Yeah, happy to go with bty = "n" too. 

(Ironically, that was my original default. But I charged it to bordered because I thought it was closer to the default plot() + legend() behaviour. Was likely overthinking it...)

I'll change both once @vincentarelbundock has had a chance to weigh in.

nathaneastwood:
Right seems more natural to me

vincentarelbundock:
I don’t have a view on this.

PR #10, opened 2023-04-09
[WIP] by_col + by_pch

vincentarelbundock:
This PR modularizes things a bit to avoid over-long functions, and allows the use of `col` to pick individual elements of the color palette:

``` r
library(plot2)

attach(mtcars)

plot2(hp, mpg, by = cyl, col = 1:4)
# Error: `col` must be of length 1 or 3.

plot2(hp, mpg, by = cyl, col = c(2, 4, 7))
```


grantmcdermott:
This looks(ed) great @vincentarelbundock. I guess you're deleting to avoid the git merge sprawl?

Thanks too for adding tests! I've been meaning to get around to that. Two quick points on that score:

1. HU that I'll be changing that default legend args per #7 at some point, hopefully today. Might be worth being explicit with `legend.position = ...` and `legend.args(bty = ...)` for any test images in the interim.
2. Should we set the OS as Linux and depend on tinysnapshot > 0.2.0? (Unfortunately, I've noticed my Arch system sets slightly different plots to Ubuntu. So I might need to set up a Rocker imagine for local testing.)

vincentarelbundock:
Yeah, I broke the merge and started fresh here: https://github.com/grantmcdermott/plot2/pull/11

In general, I think it's a good idea to just leave the defaults instead of setting `legend.position`. Forces you to make a conscious decision when defaults change. To generate new snapshots, all you have to do is:

1. Delete files in `inst/tinytest/_tinysnapshot`
2. Run: `tinytest::run_test_dir("inst/tinytest")`

Arrgh! (ubuntu vs arch)

Maybe easiest is to generate snapshots on the machine you use most often.

PR #12, opened 2023-04-10
Idea: Handle plots with factors by introducing mfrow/facets

zeileis:
**Disclaimer:** This is not a finished PR, yet, but just a proof of concept.

**Idea:** In the case where at least `x` or `y` is a factor, it is not easily possible to handle `by` within the same panel. So a simple option would be to resort to multiple facets in that case. For now I have named the corresponding argument `mfrow`, following the base `par()` option.

**Examples:**

```
plot2(Petal.Length ~ Petal.Width | Species, data = iris)               ## multiple groups in one facet
plot2(Petal.Length ~ Petal.Width | Species, data = iris, mfrow = TRUE) ## multiple facets
```

But also support for factor variables, e.g., for boxplots in one or more facets.

```
data("whiteside", package = "MASS")
plot2(Gas ~ Insul, data = whiteside)
plot2(Gas ~ Insul | Temp > 4, data = whiteside)
whiteside |>
  transform(Temp = cut(Temp, c(-1, 3, 7, 11))) |>
  plot2(x = Gas ~ Insul | Temp)
```

Or even facets of mosaic displays (aka doubledecker plots) if all variables are factors.

```
data("Titanic", package = "datasets")
ttnc <- as.data.frame(Titanic)
ttnc <- ttnc[rep(1:nrow(ttnc), ttnc$Freq), 1:4]
names(ttnc)[2] <- "Gender"
ttnc <- transform(ttnc, Treatment = factor(
  Gender == "Female" | Age == "Child", levels = c(FALSE, TRUE),
  labels = c("Normal\nMale&Adult", "Preferential\nFemale|Child")))
plot2(Survived ~ Treatment | Class, data = ttnc, mfrow = c(1, 4))
```

**To do:** Provided you think this is worth pursuing, I would still have to work out how to set all defaults properly and which graphical parameters to pass on in which specification etc.


grantmcdermott:
Oh, this is great.

A faceting option for groups has been on my to to-do list for a while. The idea that the same infra could be used to tackle factors is a cherry on top. 

(As an aside, I love the simplicity of the `mfrow = TRUE` argument. I'd been trying to think of a good name in my wishlisting and it's so obvious now that you've written it.)

As you say, the ideal functionality would provide a clean aesthetic. I think we can reduce the internal margins (`mar`) between plots and obviously use group variable names for the facets. And `mtext` for overall x and y labels. The trickier thing would be deciding on duplication of axes labels, ticks, etc within the individual facets. That, in turn, may ultimately depend on whether users want to have the same scale for each plot. (Something similar to `facet_wrap(~grp, scales = "free")` in ggplot2.) But maybe that's not something to get bogged down on during this first iteration.

In summary: I love it this idea and happy to provide more explicit feedback once you hit any deeper design decisions. Thanks for taking the lead @zeileis!

zeileis:
Great, thanks for the quick feedback, Grant @grantmcdermott!

The next question is then how extensively we want to support the faceted case. My current implementation is very lightweight, calling `plot.formula()`. But if we want full control over axes etc. we probably need to re-implement various details. This is not very difficult but a design decision we need to take. If we move in this direction, then we should probably split up `plot2.default()` and have helper functions `plot2_num_num()`, `plot2_cat_num()`, etc. These will need to have somewhat different arguments for controlling the appearance, e.g., the spinogram needs a `breaks` argument for splitting the x-axis into intervals.

To start with this, I could implement the three other `plot2_x_y()` that are not `plot2_num_num()`. And then we could start factoring out the latter after that. What do you think?

grantmcdermott:
> To start with this, I could implement the three other `plot2_x_y()` that are not `plot2_num_num()`. And then we could start factoring out the latter after that. What do you think?

I'm all for it. 

My main concern is that I don't want this project to become an additional burden for you and @vincentarelbundock. I know you are both as busy as I am, so don't want to create a weight of expectation from my side. OTOH I do think think **plot2** has the scope to become a useful and widely used package. If you have time and motivation, I 'm very excited/grateful to have these additions.

PS. I think increasing modularization is generally a good idea for this project. The current structure is very much a legacy of the original, narrowly defined script that I had lying around. Once we have implemented things like `plot_x_y()` and co., then it should also get easier to extend other "base" graphics function like grouped density plots or loess curves, etc. (Assuming ofc that we deem these worth the effort and within spec.)

zeileis:
Argh, I tried to merge the 'mfrow' branch with 'main' before continuing but somehow exceeded my git skills (I'm still not very good at this...). So I decided to close this PR and I'll start fresh. Sorry!

Issue #4, opened 2023-04-08
Known issues


grantmcdermott:
- [x] #16
- [x] "col" argument could be smarter (e.g., inherit from by) [Closing b/c I think we have a better solution with #5 now.]
- [x] #27
- [x] #26
- [x] #17
- [x] #65

zeileis:
Regarding the last question: I think it's fine if pch does not vary by group by default. But it would be nice to support something like `plot(y ~ x | z, pch = c(1, 2))` if `z` is a factor with two groups. Similarly, `col = c(3, 5)` if you want other colors than the first two from the current `palette`.

Example: It would be nice if the `plot2()` below would yield the same kind of plot as the `plot()`.

```
data("whiteside", package = "MASS")
plot(Gas ~ Temp, data = whiteside, col = palette.colors()[2:3][Insul], pch = c(19, 17)[Insul])
plot2(Gas ~ Temp | Insul, data = whiteside, col = 2:3, pch = c(19, 17))
```

etiennebacher:
Hello, I rarely use base plots in R, mostly because I don't find it very convenient so thank you for this package!

It looks like in `plot.default()`, the argument `log` is passed to `plot.window()` (which then uses some internal C code), so isn't it enough to pass this argument to `plot.window()` in `plot2()`? Or am I missing something?

I can try a PR for this if you want

grantmcdermott:
> so isn't it enough to pass this argument to plot.window() in plot2()

It could well be... I just never got around to trying or testing. A PR would be very welcome!

PR #15, opened 2023-04-11
Add argument `log`

etiennebacher:
As I said in #4, I'm not an expert with base R plots but it seems to work fine:

``` r
library(plot2)

# from ?plot.default
lx <- seq(1, 5, length.out = 41)
yl <- expression(e^{-frac(1,2) * {log[10](x)}^2})
y <- exp(-.5*lx^2)

plot(10^lx, y, log = "xy", type = "l", col = "purple",
     main = "Log-Log plot", ylab = yl, xlab = "x")
```

``` r

plot2(10^lx, y, log = "xy", type = "l", col = "purple",
     main = "Log-Log plot", ylab = yl, xlab = "x")
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -43,7 +43,7 @@ Suggests:
   tinysnapshot,
   tinytest
 Encoding: UTF-8
-RoxygenNote: 7.2.3
+RoxygenNote: 7.2.3.9000
 URL: https://grantmcdermott.com/plot2/,
     http://grantmcdermott.com/plot2/
 BugReports: https://github.com/grantmcdermott/plot2/issues
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # plot2 0.0.2.9000 (development version)
 
+New features:
+
+- Add support for the argument `log` (#15, @etiennebacher).
+
 # plot2 0.0.2
 
 Breaking changes:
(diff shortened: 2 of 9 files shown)


grantmcdermott:
Excellent! Let's see how the tests shake out. (There's a bit of a tricky situation in that we need the plots to have been generated from a Linux instance, which you hopefully are.)

Could I also request an extra test using a logged axis (axes) and the `by` (or `|`) argument? Just want to make sure there isn't some unexpected interaction with grouped plots.

etiennebacher:
> Looks like you need to update the docs (devtools::document() or similar).

Ah right, thanks

> from a Linux instance, which you hopefully are.

Yes I have Ubuntu 22.10

> Could I also request an extra test using a logged axis (axes) and the by (or |) argument?

Hum turns out it's trickier than I thought. I can't even put two `plot2()` side-by-side. I'll have a look tomorrow but feel free to jump into this one if you have time ;)

grantmcdermott:
> Hum turns out it's trickier than I thought. I can't even put two plot2() side-by-side.

That's an internal restriction. We can't use `mfrow` on grouped plot2's, because we need to alter the `par`'s internally for legend adjustment etc. I'm hoping to fix/accommodate this in a future iteration. But for now, just feel free to test a single grouped plot a la `plot2(log(Temp) ~ Day | Month, airquality)`.

grantmcdermott:
Closes #15.

grantmcdermott:
Will add another test separately. Thanks @etiennebacher!

PR #23, opened 2023-04-14
tinytest 0.0.3 is now on CRAN

vincentarelbundock:
No longer need it in `Remotes:`


grantmcdermott:
@vincentarelbundock Do you mean tinysnapshot instead of tinytest? If so, shouldn't the DESCRIPTION version requirement be on the former?

vincentarelbundock:
Haha, yes of course. I have now updated the file.

Issue #19, opened 2023-04-12
API change: single `legend` and `palette` arguments


grantmcdermott:
*TL;DR* Roll the `legend.position` and `legend.args` arguments into a single `legend = legend(...) / legend = <keyword>` argument. Similarly, roll the `palette` and `palette.args` arguments into a single `palette = <function> / palette = <keyword>` argument.

### Problem

An API/design decision I'd like to nail down before a CRAN submission is the best way to handle the `legend` and `palette` arguments. At the moment, each of these is broken into two user-facing arguments. Respectively:

- **legend:** `legend.position` (accepts a keyword string like "right!" or "topleft") and `legend.args` (accepts a list of arguments, passed down to `legend()`)
- **palette:** `palette` (accepts a string for like "Viridis" that matches at least one output from `palette.pals()` or `hcl.pals()`) and `palette.args` (accepts a list of arguments, passed down to `palette.colors()` or `hcl.colors()`).

The current setup has some advantages, but I don't really like that we require two separate arguments for each "aspect". Other concerns include the fact that we don't use all arguments available to, say, legend creation and existing arguments don't necessarily align with their base equivalents. (An example: the "title" argument in `legend.args` is equivalent to the "legend" argument in the base `legend()` function.) Finally, the current API also places some constraints on user-specified inputs. For example, if a user wanted to supply a bespoke colour-generating palette as a function.

### Proposal

An alternative would be to roll each of these into a single argument that accepts a function (and possibly a string for a convenient shorthand to change a leading/key feature). Specifically:

- **legend:** would accept a regular legend function, e.g. `legend = legend(bty = "o", legend = "my legend title")`, where we splice in any remaining defaults. We could also catch/allow users to pass a key position string as an alternative convenience option, e.g. "bottom!". (One slight concern/inconsistency is whether the keyword behaviour should be reserved for position, as I've suggested here, or whether it should be the legend title...)
- **palette:** would accept a palette-producing function like `palette()`, or `palette.cols()`, `hcl.colors()`. The trick here is that the leading "n" argument would  usually (always?) be redundant, since we calculate n internally based on the number of groups in `by`. Similar to the current behaviour, we could catch/allow users to pass down a known palette string as a convenience option e.g., "Tableau 10".

### Decision?

Personally I'm leaning towards the proposed setup, despite be a breaking change. I would love to hear other opinions before putting in a PR, though.

vincentarelbundock:
[...]

Issue #2, opened 2023-04-07
Wishlist: Support factors as x or y


zeileis:
**Thanks:**

Grant @grantmcdermott, the project looks really nice and useful, thanks!

**Wishlist:**

I just wanted to put on the wishlist that in addition to numeric `x` and `y`, it would be great if the factor-based `plot()` flavors were also supported in `plot2()`. Especially the plots where `y` is a factor are in my opinion underappreciated in base R. (It's not surprising that I would say that because I implemented these...)

**Plots:**

In `plot(x, y, ...)` and `plot(y ~ x, ...)` we have the following:

| | `x`<br/>Numeric | &nbsp;<br/>Factor |
|-----:|:-----|:-----|
|**`y`** &nbsp; **Numeric** | Scatterplot | Parallel boxplots |
| **Factor** | Spinogram <br/> (Spineplot with histogram-style breaks in `x`) | Spineplot <br/> (Flavor of mosaic plot) |
 
**Examples:**

```
data("SwissLabor", package = "AER")
plot(income ~ age, data = SwissLabor)            ## Numeric ~ Numeric
plot(income ~ foreign, data = SwissLabor)        ## Numeric ~ Factor
plot(participation ~ age, data = SwissLabor)     ## Factor ~ Numeric
plot(participation ~ foreign, data = SwissLabor) ## Factor ~ Factor
```

It would be great if the plots above would work with `plot2()` as well - and if the coloring and grouping etc. would also be supported.

_GM edit: adding checklist to track_

- [x] Numeric ~ Numeric
- [x] Numeric ~ Factor (#154)
- [ ] Factor ~ Numeric
- [ ] Factor ~ Factor

grantmcdermott:
Thanks Achim.

I can't promise I'll get to this soon, but I just invited you to the project. No pressure (I'm well aware of how many other plates you have in the air...), but think of it as in invitation to take a stab at these if you get a chance ;-)

zeileis:
Thanks! I'll try...

zeileis:
**Idea:**

I've taken another look at the current state of the code and my idea to tackle support of factor variables and also of facets is the following:

* We set up internal functions that draw some type/class of `x` against some type/class of `y`, potentially also supporting a `by` grouping variable.
* Multiple functions might be available for the same combination of `x` and `y`, e.g., `cdplot` vs. `spineplot` in case of a factor `y` and a numeric `x`.
[...]

PR #30, opened 2023-04-18
Unified palette arg

grantmcdermott:
Addresses the second part of #19 .

Some quick notes:

- I've left the default colour palette as Okabe-Ito for the moment to becnhmark against our existing test suite.
- I had to do some tricksy `substitute` footwork to get all eventualities covered, i.e. users can pass strings or functions, where we can still splice in the number of colours (groups) later on. Similarly, I also had to drop `palette` as an argument for `plot2.formula` and `plot2.density` and rather pass this through `...`. Otherwise `substitute` doesn’t work properly. Maybe there's a more failsafe way to do this?

Examples:

``` r
library(plot2)

plot2(
  Temp ~ Day | Month, airquality,
  type = "b", pch = 16
  )
```

``` r
plot2(
  Temp ~ Day | Month, airquality,
  type = "b", pch = 16,
  palette = "tableau 10"
  )
```

``` r
plot2(
  Temp ~ Day | Month, airquality,
  type = "b", pch = 16,
  palette = palette.colors(palette = "Tableau 10", alpha = 0.5)
  )
```

``` r
plot2(
  Temp ~ Day | Month, airquality,
  type = "b", pch = 16,
  palette = khroma::color("muted")
  )
```

<sup>Created on 2023-04-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/by_aesthetics.R
+++ b/R/by_aesthetics.R
@@ -1,58 +1,70 @@
-by_col = function(ngrps, col = NULL, palette = NULL, palette.args = NULL) {
+by_col = function(ngrps = 1L, col = NULL, palette = NULL) {
   
-  if (is.null(col)) {
+  # palette = substitute(palette, env = parent.env(environment()))
+  
+  if (is.null(col) && is.null(palette)) {
     col = seq_len(ngrps)
   }
-
+  
   if (is.atomic(col) && is.vector(col)) {
     if (length(col) == 1) {
-        col = rep(col, ngrps)
+      col = rep(col, ngrps)
     } else if (length(col) != ngrps) {
-        stop(sprintf("`col` must be of length 1 or %s.", ngrps), call. = FALSE)
+      stop(sprintf("`col` must be of length 1 or %s.", ngrps), call. = FALSE)
     }
-    if (is.character(col)) return(col)
-  }
-
-  if (is.null(palette)) {
-    if (ngrps<=9) {
-      palette = "Okabe-Ito"
-      palette_fun = palette.colors
-    } else {
-      palette = "Viridis"
-      palette_fun = hcl.colors
-    }
-  } else if (palette %in% palette.pals()) {
-    palette_fun = palette.colors
-  } else if (palette %in% hcl.pals()) {
-    palette_fun = hcl.colors
-  } else {
-    warning(
-      "\nPalette string not recogized. Must be a value produced by either",
-      "`palette.pals()` or `hcl.pals()`.",
-      "\nUsing default option instead.\n",
-      call. = FALSE
-      )
-    if (ngrps <= 9) {
-      palette = "Okabe-Ito"
-      palette_fun = palette.colors
-    } else {
-      palette = "Viridis"
-      palette_fun = hcl.colors
+    if (is.character(col)) {
+      return(col)
     }
   }
-
-  # n is a required argument for viridis and other palettes
-  if (!"n" %in% names(palette.args)) palette.args[["n"]] = max(col)
-
-  out = do.call(
-    function(...) Map(palette_fun, palette = palette, ...), 
-    args = palette.args
-    )[[1]]
   
-  out = out[col]
- 
-  return(out)
-
+  if (is.null(palette)) {
+    
+    if (ngrps<=8) {
+      palette = "Okabe-Ito" #"R4"
+      palette_fun = palette.colors
+    } else {
+      palette = "Viridis"
+      palette_fun = hcl.colors
+    }
+    args = list(n = ngrps, palette = palette)
+    
+  } else {
+    
+    if (class(palette)=="character") {
+      if (tolower(palette) %in% tolower(palette.pals())) {
+        palette_fun = palette.colors
+      } else if (tolower(palette) %in% tolower(hcl.pals())) {
+        palette_fun = hcl.colors
+      } else {
+        stop(
+          "\nPalette string not recogized. Must be a value produced by either",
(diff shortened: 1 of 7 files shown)


zeileis:
Looks good to me. I just saw the `is.character()` issue that you already fixed and that `"R4"` is still commented.

_(Disclaimer: I just read the code and didn't run it.)_

grantmcdermott:
Thanks @zeileis! Sorry, I was still busy adding additional context (and examples) before you responded. Everything seems to be running pretty well from my side. But please do kick the tyres and give the code some scrutiny.

zeileis:
For the tricksy stuff and splicing in `n`: Personally, I find this confusing and wouldn't use it. But others will probably perceive this differently.

An alternative would be to simply allow the specification of more colors than `ngrps` and then just selecting `ngrps` colors from it:

* From something like `palette.colors(palette = "Okabe-Ito", alpha = 0.5)` you always get the maximum number of colors in the corresponding palette and you can simply select the first `ngrps` from it.
* The same is not possible for `hcl.colors()` but instead you could select a large number of colors and then select an equidistant subset from it or go via `colorRampPalette()` to interpolate the colors.
* Then only the question remains how to distinguish the case where you select the first `ngrps` colors and where you interpolate. One strategy could be to switch to the latter if the number of colors is greater than 36 because that is the largest qualitative palette I'm aware of. Also not ideal and potentially confusing but less "magical" than the evaluation tricks in my opinion.

grantmcdermott:
Super, thanks for the feedback @zeileis. I hope to circle to this once I get some other things out of the queue. Let's hope Vincent finds a sec to take a look too, since I'm overwriting his `by_col` function.

vincentarelbundock:
This looks like an excellent change! I wonder if it might not make sense to do some pre-processing on the palette argument before passing it to `by_col`. Passing through `...` seems not great, so it might be worth having a bit of boilerplate duplication for a cleaner interface.

Sorry I don't have to look at this in more detail now.

zeileis:
Not much of an update here, I just wanted to show a brief illustration of what I thought could be a useful strategy for reducing a longer color vector to the `ngrps` needed.

For the qualitative palettes from `palette.colors()` we can simply select the first `ngrps` (below set to 5 for illustration). The these two are identical:

```
palette.colors(n = 5, palette = "Okabe-Ito")
##       black      orange     skyblue bluishgreen      yellow 
##   "#000000"   "#E69F00"   "#56B4E9"   "#009E73"   "#F0E442" 
palette.colors(palette = "Okabe-Ito")[1:5]
##       black      orange     skyblue bluishgreen      yellow 
##   "#000000"   "#E69F00"   "#56B4E9"   "#009E73"   "#F0E442" 
```

For `hcl.colors()` we can reduce the number of colors to `ngrps` (set to 12 for illustration below) by going through `colorRampPalette()`. This will yield very similar or in some cases even idential output to setting `n = ...` explicitly.

```
hcl.colors(12, palette = "Viridis")
##  [1] "#4B0055" "#45256B" "#30437F" "#005F8E" "#007896" "#009097" "#00A691"
##  [8] "#00B983" "#3AC96D" "#8BD650" "#C8DF32" "#FDE333"
colorRampPalette(hcl.colors(100, palette = "Viridis"))(12)
##  [1] "#4B0055" "#45256B" "#30437F" "#005E8E" "#007896" "#009097" "#00A591"
##  [8] "#00B883" "#3AC96D" "#8BD650" "#C8DF32" "#FDE333"
```

There are two caveats: First, `colorRampPalette()` drops the `alpha` channel so this would have to be re-added (if used). Second, we somehow need to distinguish the two cases. One heuristic could be the length of the color vector (as suggested in my previous comment). An alternative heuristic would be to set an attribute to the color vector so that we remember which function was used for generating the palette. Then we would know how to handle `palette.colors` and `hcl.colors` and only need to decide which strategy to use otherwise.

grantmcdermott:
(Just a quick note to apologize for being slow with this: I was hoping to get to your comments during the week, but I had too much else going on. I'm heading out on vacation tomorrow, but will be back next week to finish this off.)

vincentarelbundock:
Vacation should always be priority 1. Enjoy!

grantmcdermott:
Hi folks. Again, sorry for keeping you both waiting on this.

So, after of thought, I've decided to merge this PR as-is. I don't mean to fob off the legitimate points and concerns that you have raised @zeileis (far from from it!). But I think there are two arguments for expediency:

1. The current implementation gives us _user-level_ functionality that is close to optimal IMHO.  I'm very happy for us to take another crack at the internals once this PR is merged, so please feel free to have a crack it afterwards. On that note, Achim I like your `colorRampPalette` example/idea a lot and you're certainly better placed than am I to make the code work. One UI thing I'd like to avoid, however, is requiring users to guess `n` ahead of time, or (similarly) having to supply a large `n` as a placeholder that will ultimately get overwritten.
2.  The other reason is that I don't want to keep this PR from dragging on is to unblock @zeileis from the faceting / factor modules in #2. I'll ping that thread as soon as I'm done with some other little changes I want to make.

zeileis:
Fair enough. I would still argue, though, that stopping if `n` does not match is not very user-friendly:

https://github.com/grantmcdermott/plot2/blob/main/R/by_aesthetics.R#L12-L14

And using `colorRampPalette()` for going from a palette with a fixed set of colors to a palette with the desired number of colors is something that is quite common, I think. This is what is done for all the palette from `RColorBrewer` because they just provide a fixed set anyways.

But as you said: We can also revisit this later.

Issue #29, opened 2023-04-17
Should we support `symbols(2)`?


grantmcdermott:
Mostly for bubble charts. You can imagine the background colour below being assigned by group (as opposed to random recycling here).

``` r
with(
    data.frame(state.x77),
    {
        symbols(
            x = Income, y = Life.Exp,
            circ = sqrt(Population / pi),
            inches = .25,
            fg = "white",
            bg = palette.colors(n = 4, alpha = 0.5)[-1],
            main = "Life expectancy among US states (1977)"
        )
    }
)
```

![](https://i.imgur.com/gjI4eqP.png)

<sup>Created on 2023-04-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

grantmcdermott:
Upon reflection, it's probably better (easier) from a maintenance perspective just to do this via `cex`. Here, using a similar rescaling factor to ggplot2.

``` r
rscl = function (x) {
    from = range(x)
    to = c(1, 6)
    (x - from[1])/diff(from) * diff(to) + to[1]
}

with(
    data.frame(state.x77),
    plot(
        x = Income, y = Life.Exp,
        pch = 21,
        cex = rscl(Population),
        col = "white",
        bg = palette.colors(n = 4, alpha = 0.5)[-1],
        main = "Life expectancy among US states (1977)"
    )
)
```

![](https://i.imgur.com/2CtLm99.png)

<sup>Created on 2023-07-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

grantmcdermott:
Closing in favour of #48.

PR #50, opened 2023-07-28
Support cex and bg

### LINKED ISSUE #48: cex arg not working
Compare:

``` r
plot(0:10, cex = 3, main = "plot")
```

``` r

library(plot2)
plot2(0:10, cex = 3, main = "plot2")
```

<sup>Created on 2023-07-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Same for `cex.main`, `cex.sub`, and `cex.axis`, although these probably require separate attention via `...` passing.

### LINKED ISSUE #49: bg arg not working
Compare:

``` r
plot(0:10, pch = 21, col = "black", bg = "red", main = "plot")
```

``` r

library(plot2)
plot2(0:10, pch = 21, col = "black", bg = "red", main = "plot2")
```

<sup>Created on 2023-07-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

We probably want to support `bg = "by"` too, so that we can support something like `plot2(..., col = "white", bg = "by")`.

grantmcdermott:
Closes #48. Closes #49.

In addition to specifying these parameters manually, this PR also permits `bg = "by"` (similar to existing "by" keyword support for `lty` and `pch`).

I haven't added `cex = "by"` keword support yet, because I need to solve the potential problem of dual legend placement, e.g. one legend for colors and another for cex/size. However, once that's been cracked then it should be fairly trivial to support "bubble" plot types (c.f., #29).

Quick example of what this PR allows:

``` r
library(plot2)

palette("Tableau 10")
palette(adjustcolor(palette() , 0.3))

plot2(
    Sepal.Length ~ Petal.Length | Species, iris,
    pch = 21,
    col = "black",
    bg = "by",
    cex = 2
)
```

<sup>Created on 2023-07-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/plot2.R
+++ b/R/plot2.R
@@ -338,7 +338,6 @@ plot2.default = function(
   } else {
     bg = rep(bg, ngrps)
   }
-  cat("\n", bg, "\n")
   
   # Save current graphical parameters
   opar = par(no.readonly = TRUE)
--- a/R/plot2.R
+++ b/R/plot2.R
@@ -90,20 +90,30 @@
 #'    specifying "none").
 #'    - A list or, equivalently, a dedicated `legend()` function with supported
 #'    legend arguments, e.g. "bty", "horiz", and so forth.
+#' @param col plotting color. Character, integer, or vector of length equal to
+#'   the number of categories in the `by` variable. See `col`. Specifying colors
+#'   manually in `plot2` should not be necessary unless users wish to override
+#'   the automatic colors produced by the grouping process. Typically, this
+#'   would only be done if grouping characterists are deferred to some other
+#'   graphical parameter (i.e., passing the "by" keyword to one of `pch`, `lty`,
+#'   or `bg`; see below.)
 #' @param pch plotting "character", i.e., symbol to use. Character, integer, or
 #'   vector of length equal to the number of categories in the `by` variable.
 #'   See `pch`. In addition, users can supply a special `pch = "by"` convenience
 #'   argument, in which case the characters will automatically loop over the
 #'   number groups. This automatic looping will begin at the global character
 #'   value (i.e., `par("pch")`) and recycle as necessary.
-#' @param col plotting color. Character, integer, or vector of length equal to
-#'   the number of categories in the `by` variable. See `col`.
 #' @param lty line type. Character, integer, or vector of length equal to the
 #'   number of categories in the `by` variable. See `lty`. In addition, users
 #'   can supply a special `lty = "by"` convenience argument, in which case the
 #'   line type will automatically loop over the number groups. This automatic
 #'   looping will begin at the global line type value (i.e., `par("lty")`) and
 #'   recycle as necessary.
+#' @param bg background (fill) color for the open plot symbols 21:25: see
+#'   `points.default`. In addition, users can supply a special `bg = "by"`
+#'   convenience argument, in which case the background color will inherit the
+#'   automatic group coloring intended for the `col` parameter.
+#' @param cex character expansion.
 #' @param par_restore a logical value indicating whether the `par` settings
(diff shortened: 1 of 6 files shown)


vincentarelbundock:
Looks fantastic! Well done!

Two examples:

``` r
library(plot2)

palette("Tableau 10")
palette(adjustcolor(palette() , 0.3))

# Should `col` accept "by"?
plot2(
    Sepal.Length ~ Petal.Length | Species, iris,
    pch = 21,
    col = "by",
)
# Error in plot.xy(xy.coords(x, y), type = type, ...): invalid color name 'by'

# Default looks really pale on my screen
plot2(Petal.Length ~ Sepal.Length | Species, iris)
```

R/plot2.R
```diff
@@ -256,8 +266,10 @@ plot2.default = function(
     palette = NULL,
     legend = NULL,
     pch = NULL,
-    col = NULL,
     lty = NULL,
+    col = NULL,
+    bg = NA,
```
vincentarelbundock, line 271: Should default be `NULL` like for the other arguments?

R/plot2.R
```diff
@@ -256,8 +266,10 @@ plot2.default = function(
     palette = NULL,
     legend = NULL,
     pch = NULL,
-    col = NULL,
     lty = NULL,
+    col = NULL,
```
vincentarelbundock, line 270: Should `col` accept "by"?

R/plot2.R
```diff
@@ -90,20 +90,30 @@
...
 #'   number of categories in the `by` variable. See `lty`. In addition, users
 #'   can supply a special `lty = "by"` convenience argument, in which case the
 #'   line type will automatically loop over the number groups. This automatic
 #'   looping will begin at the global line type value (i.e., `par("lty")`) and
 #'   recycle as necessary.
+#' @param bg background (fill) color for the open plot symbols 21:25: see
+#'   `points.default`. In addition, users can supply a special `bg = "by"`
+#'   convenience argument, in which case the background color will inherit the
+#'   automatic group coloring intended for the `col` parameter.
+#' @param cex character expansion.
```
vincentarelbundock, line 116: docstring could be more explicit. Here is what `?par` says: a numerical vector giving the amount by which plotting characters and symbols should be scaled relative to the default. This works as a multiple of par. NULL and NA are equivalent to 1.0. Note that this does not affect annotation: see below.

R/plot2.R
```diff
@@ -316,7 +328,16 @@ plot2.default = function(
     ngrps = ngrps,
     col = col,
     palette = substitute(palette)
+  )
+  if (!is.na(bg) && bg == "by") {
```
vincentarelbundock, line 332: Nice re-use!

vincentarelbundock:
Looks really great overall. I left some comments. Once you've made decisions on those, I won't have anything else to add and think it's good to merge.

R/plot2.R
grantmcdermott: Good point. I just took from the `plot.xy` defaults, but we should rather strive to be internally consistent here.

R/plot2.R
grantmcdermott: It's a good question (and probably answered with "yes"). Initially, the default (and only) behaviour was for groups to be represented through colour. So `col="by"`was kind of redundant even after we added support for other graphical parameters. But ofc that doesn't mean it should error if users _do_ specify it manually.

grantmcdermott:
Super, thanks @vincentarelbundock.

Minor aside RE the paleness of this palette. That should be resolved by selecting a solid point type (e.g., `pch = 19`). Or, by keeping the filled point type (`pch = 21`) but explicitly selecting some fill (e.g. bg = "by" / bg = "grey", etc.)

PR #51, opened 2023-08-02
Fix y-title for function names

etiennebacher:
Close #47. I didn't try on many plots myself but it looks like it passes the CI. Sorry for all for the git changes, I don't know why it's like this, the changes are just a few lines at the beginning of `plot2.default()`.

@grantmcdermott I'm still struggling with the tests suite of this package (esp. because I'm currently on Windows) so it's probably better if you can take over this PR to generate snapshots. 

New behavior:

``` r
set.seed(42)
plot(rnorm(1))
```

``` r

library(plot2)
set.seed(42)
plot2(rnorm(1))
```

PR #52, opened 2023-08-02
 Fix y-title for function names 

### LINKED ISSUE #47: y-title for function names
Compare:

``` r
set.seed(42)
plot(rnorm(1))
```

``` r

library(plot2)
set.seed(42)
plot2(rnorm(1))
```

<sup>Created on 2023-07-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

etiennebacher:
Same as #51, close #47

Much better like this @grantmcdermott

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,9 +23,15 @@ Authors@R:
       role = "aut",
       email = "Achim.Zeileis@R-project.org",
       comment = c(ORCID = "0000-0003-0918-3766")
-     )
+     ),
+    person(
+      given = "Etienne",
+      family = "Bacher",
+      role = "ctb",
+      email = "etienne.bacher@protonmail.com"
+    )
   )
-Description: Lightweight extension of the base R plot function, with support for
+Description: Lightweight extension of the base R plot system, with support for
       automatic grouping and legend printing, and several other enhancements.
 License: GPL (>= 2) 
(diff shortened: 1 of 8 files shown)


etiennebacher:
What? No idea why it fails now

grantmcdermott:
Super, thanks. I'll take a look on my local Linux machine as soon as the kids get to school.

grantmcdermott:
Hmmm. It looks like the Hershey font tests (and one par restore test) are triggering the CI errors. It's almost certainly nothing to do with this PR, but I'd like to test a bit more locally. I have to jump on some other work first, but will circle back to this later.

etiennebacher:
This is only on R-devel, could it be due to a recent commit there? For example https://github.com/wch/r-source/commit/fa69bab0e61ceca6d186afe7c91f2d2b558db718?

grantmcdermott:
> This fails only on R-devel, could it be due to a recent commit there? 

Thanks, I think that must be it. I downloaded the action artifact here and manually inspected the `_tinysnapshot_review` diffs. I honestly can't see what the difference is.... maybe some incredibly slight shading/alpha change? I'll wait for R 4.4.0 to re-generate these snapshots locally

At any rate, happy to accept the PR now. Thanks for the contribution @etiennebacher !

PR #62, opened 2023-08-15
Basic one-sided formula support

grantmcdermott:
quick examples:

``` r
#> ℹ Loading plot2

par(pch = 16)

plot2(~ Temp, airquality)
```

``` r
plot2(~ Temp | Month, airquality, type = "b", par_restore = TRUE)
```

``` r
plot2(~ Temp | Day * Month, airquality, legend = FALSE)
```

<sup>Created on 2023-08-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: plot2
 Type: Package
 Title: Lightweight extension of base R plot
-Version: 0.0.3.9005
+Version: 0.0.3.9006
 Authors@R: 
   c(
     person(
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# plot2 0.0.3.9005 (development version)
+# plot2 0.0.3.9006 (development version)
 
 New features:
 
@@ -9,6 +9,7 @@ the fill and border. (#50 @grantmcdermott)
 - Support for filled density plots. (#58 @grantmcdermott)
 - The new `add` argument allows new plot2 objects to be added to / on top of the
 existing plot window. (#60 @grantmcdermott)
+- Support for one-sided formulas, e.g. `plot2(~ Temp | Month, airquality)`. (#62 @grantmcdermott)
 
 Bug fixes:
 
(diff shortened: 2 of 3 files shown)


grantmcdermott:
@zeileis do you mind taking a look at this? Note that this PR just provides basic one-side formula support and doesn't yet address #61. (I have some ready to roll on top of this PR that _does_ address #61, but it starts to get a little more complicated, so I want to hold back on it until we're happy with the basic scaffolding.)

zeileis:
Nice! The only problem I saw was that the `bylab` title is wrong in the case where the are multiply `by` variables. In your example you had to switch off the legend because it got too long. A more manageable example could be:

```
airq <- transform(airquality, Temp = factor(Temp > 85, levels = c(FALSE, TRUE), labels = c("low", "high")))
plot2(~ Ozone | Temp + Month, data = airq, col = rep(palette.colors(5, "Tableau"), each = 2), pch = rep(c(1, 19), 5))
```

The fix would have just been to replace `-(1L:2L)` with `-yxloc` when putting together the `bylab`. However, when I went through the code I thought that I would add some more comments, add an explicit logical variable `no_y`, simplify the error message, and a couple of other small changes. They don't change the behavior of the code but I think it is a bit more readable now. Should I push this to the branch so that you can have a look? Or should I just make the minimal change?

zeileis:
OK, I have pushed all my changes (including the cosmetic ones) now, I hope that's alright. Otherwise feel free to revert.

zeileis (approved):
With the small fix in the default `bylab` for one-sided formulas with multiple `by` variables, this is good to go, I think.

grantmcdermott:
Excellent, thanks!

Btw, on the subject of legends that are too long, I have some ideas for continuous variable legends with interpolated (gradient) palette. I’ll create an issue and tag you when I get a sec to write up my thoughts.

zeileis:
This would probably mean that we shouldn't just coerce all `by` variables to factors. The continuous gradient you mention could then be used for numeric `by` variables. However, the same idea/trick could not be used for other properties (`pch` or `lty`) that are necessarily discrete. Additionally/alternatively, one could be smarter in coercing numeric `by` variables to factors. Rather than just always using all unique variables via `factor()` one could `cut()` the variable at some quantiles - similar to what `lattice` does. My feeling was that this could easily get confusing, though, hence I did not try to do anything like that in the formula method.

Issue #63, opened 2023-08-15
Website


grantmcdermott:
In keeping with **plot2**'s goal to be an attractive lightweight plotting option, the website should help to offload content that is adding unnecessary bulk to the package itself. A case in point is the fact that all the images in the README are currently pushing the installation tarball  to several MiBs.

Serving content external to the package also means that we shouldn't use conventional R vignettes (which would get bundled with the installation tarball). So I'm not sure how well this workflow jibes with standard R website builders like **pkgdown** and **altdoc**.

Some options:

- Build a `docs/` folder locally and serve it through github-pages. (Can perhaps still be automated with GH Actions?)
- Use a Quarto book format like @vincentarelbundock did for **marginaleffects** [link](https://github.com/vincentarelbundock/marginaleffects/tree/main/book). Vincent, curious about your experiences here.
- A gallery-type website (see #59).

vincentarelbundock:
Easiest would probably be to add `vignettes/` to `.Rbuildignore`.

It was kind of hard to do a Quarto book, and I had to write a lot of custom scripts. Over the next few months, @rempsych and I will be working on improving `altdoc` in several ways, including allowing for Quarto book output.

I don't plan to use `pkgdown` for complicated packages in the future.

If you want, I can try to use `altdoc` to produce a minimalist website for `plot2` with Material MkDocs.

grantmcdermott:
> If you want, I can try to use altdoc to produce a minimalist website for plot2 with Material MkDocs.

Cool, thanks! For a MWE feel free to move the README examples to the equivalent of a vignette.

(One lingering concern I have about Material MkDocs is that this pulls in a Python dependency. Am I remembering correctly, here?)

zeileis:
We had the same problem with our [colorspace](https://CRAN.R-project.org/package=colorspace) package: We started writing quite a few vignettes which became too large for the package. So we turned them into a `pkgdown` page: <https://colorspace.R-Forge.R-project.org/>.

The _Get started_ vignette is the only vignette that is also contained in the package, all others are in `.Rbuildignore`. It has a short overview, some teaser images, and (absolute) links to the other vignettes on R-Forge. Thus, there is the same content just rendered differently on CRAN and R-Forge:

- <https://CRAN.R-project.org/web/packages/colorspace/vignettes/colorspace.html>
- <http://colorspace.R-Forge.R-project.org/articles/colorspace.html>

This setup worked quite well for us and we turned the vignettes eventually also into a paper.

[...]

PR #64, opened 2023-08-15
WIP: mkdocs & altdoc & makefile

vincentarelbundock:
Building the website requires local install of `mkdocs` and `mkdocs-material`. But of course the end-user does not need to install any of that.

I also added a Makefile with useful commands like `make deploy`. I think it should be pretty self-explanatory.

The rendered website is here: https://vincentarelbundock.github.io/plot2/


vincentarelbundock:
This PR is a first attempt to address https://github.com/grantmcdermott/plot2/issues/63

zeileis:
Disclaimer: I haven't worked with `mkdocs` or `altdoc` yet, just read about them.

Is there a way to include printed and graphical output in the examples? That's a feature that I quite like about `pkdown`.

grantmcdermott:
Nice! Thanks @vincentarelbundock , this is a great proof of concept.

I'm still quite leery of taking on a Python dependency for the website. But I do like the mkdocs look and layout.

It seems that there is still some good discussion happening at #63, so let's wait to see what comes of that.

grantmcdermott:
> Is there a way to include printed and graphical output in the examples? 

@zeileis I don't quite follow. Do you mean automatically rendered/knitted code as per a qmd or Rmd doc?

zeileis:
Yes. For illustration consider the examples at the end of:

https://vincentarelbundock.github.io/plot2/reference/

These are just shown verbatim. However, the examples from this manual page in the `pkgdown` page for `colorspace` are actually run with output/graphics included:

http://colorspace.R-Forge.R-project.org/reference/contrast_ratio.html

grantmcdermott:
Ah, got it thanks.

I think Vincent and Etienne hacked together a script for doing this with the [r-polars docs](https://rpolars.github.io/reference_home/) (but I may just be misremembering)..

vincentarelbundock:
Yep, we rolled our own script for `polars`, but rendering examples in the function documentation is not super easy to do right now.

The reason I made this PR using `altdoc` is that I've been hungry for a serious competitor to `pkgdown`, and that I plan to spend some serious time on `altdoc` in the coming months. So this was an excuse to try it out and hit the pain points in a relatively simple package (`plot2`).

So yeah, currently it doesn't have all the features we probably want eventually. But feature parity is a matter of months rather than years, I think.

vincentarelbundock:
huh, that sounded a bit dry. To be clear, I'm perfectly happy with doing `pkgdown` for this project. I was just explaining where I was coming from...

zeileis:
Thanks for the explanation, Vincent, very helpful. And also to be clear: I'm also fine with trying out something other than `pkgdown`, especially when I know that you are working on it!

I just thought that for a graphics package it would be nice to show the example plots in the manual web pages. But this is not urgent and might be an incentive to improve `altdoc`?

PR #66, opened 2023-08-16
Support type = "density"

### LINKED ISSUE #61: formula interface for densities

grantmcdermott:
Closes #61.

Supports `type = "density"` for both the atomic...

``` r
#> ℹ Loading plot2

with(airquality, plot2(Temp, type = "density"))
```

``` r
with(airquality, plot2(Temp, type = "density", by = Month, bg = "by"))
```

... and one-sided formula methods:

``` r
plot2(~Temp, airquality, type = "density")
```

``` r
plot2(~Temp | Month, airquality, type = "density", bg = "by")
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: plot2
 Type: Package
 Title: Lightweight extension of base R plot
-Version: 0.0.3.9005
+Version: 0.0.3.9006
 Authors@R: 
   c(
     person(
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: plot2
 Type: Package
 Title: Lightweight extension of base R plot
-Version: 0.0.3.9006
+Version: 0.0.3.9007
 Authors@R: 
   c(
     person(
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# plot2 0.0.3.9006 (development version)
+# plot2 0.0.3.9007 (development version)
 
 New features:
 
@@ -11,6 +11,9 @@ the fill and border. (#50 @grantmcdermott)
 existing plot window. (#60 @grantmcdermott)
 - Support for one-sided formulas, e.g. `plot2(~ Temp | Month, airquality)`. (#62
 @grantmcdermott and @zeileis)
+- Support for `plot2(x, type = "density")` as an alternative to
+`plot2(density(x))`. Works for both the atomic and one-sided formula methods.
+(#66 @grantmcdermott)
 
 Bug fixes:
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# plot2 0.0.3.9005 (development version)
+# plot2 0.0.3.9006 (development version)
 
 New features:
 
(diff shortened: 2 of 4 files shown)


grantmcdermott:
@zeileis Do you mind looking over this as the follow-up to #62? I had to do a bit more work internally to make sure that everything gets passed around correctly. Please feel free to push changes / make suggestions if you spot room for simplification. Thanks!

vincentarelbundock:
No constructive comment, but I really love these, especially with the white lines (black should still be default, obv).
white

grantmcdermott:
> No constructive comment, but I really love these, especially with the white lines (black should still be default, obv).

Yeah, messing around with `bg` and `col` in these filled density plots is unreasonably pleasing to me. (Another example [here](https://github.com/grantmcdermott/plot2/pull/58#issue-1847817449).)

grantmcdermott:
(One side issue I'd like feedback on: Should we try to emulate the automatic main title of `plot.density` as I've got here? Or, should we just treat type = "density" as a special case of regular plot and so no title by default?)

vincentarelbundock:
The second option (no default) sounds like it would be (a) more consistent, (b) not as automagical, and (c) less work. Win-win-win?

zeileis (commented):
Thanks, this mostly looks good to me. Just a couple of questions about:

https://github.com/grantmcdermott/plot2/pull/66/files#diff-ac92c4aae42f7ec2cea785da46c64d2ac425937582fb17a253fec82f4fba6e55R318-R324

I agree with Vincent that I would rather be consistent with other `plot2()` plots rather than with the base `plot.density`.

Why does is the check for `is.null(y)` done here? Shouldn't this be checked more generally for the density (and not just for `main`)?

If we omit the title, this is not relevant: But I think we cannot be sure that `density.default()` is called. This will usually be the case, of course, because it is the only `density()` method in `graphics` but we do call the generic, not the default method.

grantmcdermott:
Thanks both. I pushed some small changes to the repo so that we don't produce a title now.

I have to specify `main = NA` internally (after first checking for an explicit user-provided title ofc). Otherwise, we'll get still get an automatic and misleading `default.density(x = x)` title once it goes through `plot2.density`. But, failing that, I think that this is otherwise a pretty clean solution.

Same examples from earlier, but confirming that the title is now switched off. Just the atomic version to save space, but I've checked the formula and it's the same.

``` r
#> ℹ Loading plot2

with(airquality, plot2(Temp, type = "density"))
```



``` r
with(airquality, plot2(Temp, type = "density", by = Month, bg = "by"))
```



<sup>Created on 2023-08-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

grantmcdermott:
PS. One thing we don't do now is check / warn users that supplying a y variable argument is superfluous, e.g. `plot2(y ~ x, type = "density")`. Instead, it just gets ignored silently. I can add a message if you think it's important.

zeileis:
Nice, thanks. And, yes, I think that an explicit whether `y` is empty would be helpful.

grantmcdermott:
Done.

(We can maybe revisit this later if we think `y ~ x, type = "density"` provides a nice way to generate ridgeline density plots a la **ggridges**.)

Based on your comments, I'll assume you're happy for this to be merged now. Thanks folks!

zeileis:
Definitely happy, thanks, Grant!

PR #69, opened 2023-08-18
Support area plot types

### LINKED ISSUE #67: Support "area" plots
(Where, following ggplot2, area plots are just defined as a special case of ribbon plots with ymin = 0 and ymax = y.)

While we can manually achieve this by doing something like...

``` r
with(
    airquality,
    plot2(
        x = seq_along(Temp), 
        y = Temp,
        ymin = rep(0, length(Temp)),
        ymax = Temp,
        type = "ribbon"
    )
)
```

... it would be cool if it could be automatically inferred when (a) type = "ribbon" and (b) ymin and ymax aren't specified. So, something like the following would offer a concise way to get the previous plot above.

``` r
with(
    airquality,
    plot2(Temp, type = "ribbon")
)
```

grantmcdermott:
Closes #67.

(I also fixed a bug related to bespoke legends when passing to plot2.density via the one-sided formula method... but that's supplemental to the main purpose of the PR.)

Some quick examples.

``` r
#> ℹ Loading plot2

plot2(sin(seq(0,10, .1)), type = "area")
```

``` r

plot2(mpg ~ wt, mtcars, type = "area")
```

And then some fancier examples because we can (TM).

``` r

par(las = 1)

plot2(
    ~Temp, airquality,
    type = "area",
    col = "darkcyan",
    grid = TRUE, frame = FALSE
)
```

``` r

plot2(
    ~ Temp | Month, airquality,
    type = "area",
    grid = TRUE, frame = FALSE
)
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: plot2
 Type: Package
 Title: Lightweight extension of base R plot
-Version: 0.0.3.9007
+Version: 0.0.3.9008
 Authors@R: 
   c(
     person(
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# plot2 0.0.3.9007 (development version)
+# plot2 0.0.3.9008 (development version)
 
 New features:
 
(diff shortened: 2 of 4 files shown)


grantmcdermott:
Merging this myself as the core change ([lines 359-364](https://github.com/grantmcdermott/plot2/pull/69/files#diff-ac92c4aae42f7ec2cea785da46c64d2ac425937582fb17a253fec82f4fba6e55R359-R364)) is very simple.

vincentarelbundock:
Those examples are *killer*

PR #78, opened 2023-09-08
Handling of grid = TRUE

zeileis:
If `grid = TRUE` is specified, then the new `abline(... = pretty(extendrange(...)), ...)` is used now instead of `grid()` - and not just for date/time classes (see https://github.com/grantmcdermott/plot2/issues/77).

Reason: Unlike `grid()`, the function `pretty()` is generic. Thus, it can adapt better to axes for variables of other classes.

Only exception: For axes in logs, we still resort to using `grid()` which is likely handled better there.

Potential downside: In all cases I tried, the `pretty(extendrange(...))` approach yielded the same grid positions as `grid()`. But there may be situations when differences occur. However, in the examples I tried, I wasn't able to produce such a situation.

### DIFF
--- a/R/plot2.R
+++ b/R/plot2.R
@@ -670,11 +670,11 @@ plot2.default = function(
         if (isTRUE(grid)) {
           gnx = gny = NULL
           if (!par("xlog")) {
-            graphics::abline(v = pretty(grDevices::extendrange(x)), col = "lightgray", lty = "dotted")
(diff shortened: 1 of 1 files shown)


grantmcdermott (approved):
Nice! Good catch on the log axes and I think this helps keep cleaner internals, in addition to the generic dispatch. All checks passing, so I'm happy to merge. Users should still be able to pass `grid = grid()` if they want to enforce the "old" behaviour anyway.

Issue #79, opened 2023-10-27
Facet support


grantmcdermott:
I'm working towards basic facet support in https://github.com/grantmcdermott/plot2/tree/facets.

Quick example of what's possible so far.

``` r
devtools::load_all("~/Documents/Projects/plot2")
#> ℹ Loading plot2

data("penguins", package = "palmerpenguins")

par(pch = 19)

# basic
with(
  penguins,
  plot2(
    x = flipper_length_mm, y = body_mass_g,
    facet = species
  )
)
```

![](https://i.imgur.com/wbT0uMf.png)

``` r

# fancier version with addtional args (incl. interaction with "by" arg)
with(
  penguins,
  plot2(
    x = flipper_length_mm, y = body_mass_g,
    facet = species, by = sex,
    main = "penguins",
    grid = TRUE, frame = FALSE
  )
)
```

![](https://i.imgur.com/PkxMQT3.png)

<sup>Created on 2023-10-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Some outstanding issues that I'd like to resolve:

1. At the moment, the facet arg is specified as a standard, single atomic var. Do we instead want to enable facet support via a (potentially two-sided) formula? This might be the most natural way to enable ggplot2-style facet grid, a la `facet = rows ~  cols`... Or, do you want to support both methods somehow?
2. At the moment, facets are arranged by window columns (i.e., in a single row). Should we try to do something smarter here, like automatically fold into "square" rows and columns?
3. At the moment, the axes are printed for all facets. Do we want to only print these for the exterior facets to avoid duplication?
4. Do we want to  enable additional customization of facet options, probably via a list argument? This could include some of the things that I've mentioned above, e.g. controlling the number of cols/rows, printing of axes, etc. But I'm worried that it creates an additional maintenance burden.

PR #80, opened 2023-11-27
docsify

vincentarelbundock:
I have worked a lot on `altdoc`. It still has a few rough corners, but I've already moved all of my packages over and it looks great.

Here is what the `plot2` could look like: 

https://arelbundock.com/plot2

You'll notice that:

* Examples are evaluated and displayed in both the README and the MAN pages.
*  Unlike MkDocs, the Docsify documentation generation framework does not require additional dependencies on the dev machine. It just loads some javascript from the web.
* I changed almost nothing to the `plot2` documentation itself:
  - Fixed a typo in `plot2.Rd`.
  - Removed a `dev.off()` call which broke sequential plots in Rmarkdown.
  - Changed the headers in NEWS for better display.
  - Changed the path where README files are saved to `man/figures/README/`

Also, this is a pure **default** website. The only `altdoc` specific commands I ran were:

```r
library(altdoc)
setup_docs("docsify")
render_docs()
```


vincentarelbundock:
Last note, this is using the Dev version of `altdoc`, which is completely different from the original version. The website shows the current workflow:

https://altdoc.etiennebacher.com

vincentarelbundock:
Yeah, I think we can merge now. I split the examples into a Get Started vignette.

The basic workflow will certainly remain the same, and only some minor behind the scenes stuff should change, requiring very little adaptation here.

We may need to think about a strategy if you want a Github Actions to auto-deploy, and also if you want the separate vignette to be available on both CRAN and website (I have not fully solved that yet, but it should be easy).

For now, all you'll need to do is go to the repo settings to setup Github-Pages. It's very easy. Just do the same thing I do on countrycode. See screenshot:

Screenshot 2023-11-27 165027

zeileis:
Wow, this looks really cool, I look forward to trying it out myself for some of my other packages.

vincentarelbundock:
Glad to hear!

Maybe wait a few weeks, though . I mostly did this here as part of dogfooding process. It's not *quite* ready for public use yet.

grantmcdermott:
Okay, awesome. Please go ahead and merge when you're happy @vincentarelbundock.

I've got a PR in the works to supports faceting (#79) and once that's done then I think we can put out a new release with the shiny new website 🚀

Issue #65, opened 2023-08-15
custom `mfrow` settings are not respected for grouped plots


grantmcdermott:
Fine without a legend:

``` r
library(plot2)

par(mfrow = c(1,2))
plot2(mpg ~ wt, mtcars)
plot2(wt ~ mpg, mtcars)
```

![](https://i.imgur.com/SP2hS55.png)

Not fine with a legend:

``` r

par(mfrow = c(1,2))
plot2(mpg ~ wt | cyl, mtcars)
```

![](https://i.imgur.com/SGfJGCq.png)

``` r
plot2(wt ~ mpg | cyl, mtcars)
```

![](https://i.imgur.com/SbvqC5I.png)

<sup>Created on 2023-08-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

grantmcdermott:
Stepping through the debugger, the problem appears to stem from the various `par(omd = ...)` calls in setting up the legend, (e.g. [here](https://github.com/grantmcdermott/plot2/blob/0c7db77cf96e01cb969f8aea6c8768f91e16284a/R/plot2.R#L487)) which are somehow (?) changing `par("mfg")` at the same time. (Maybe `par("new")` too?)

I don't really understand why it's doing this and fixing may require more duct tape than is worth... especially if we're going to be passing custom `par(mfrow = ..., mfg = ...)` settings internally for facets.

zeileis:
Is it really the `par(...)`? I thought it was the `plot.new()` that causes the step to the next panel in mfrow.

grantmcdermott:
Yeah I'm really puzzled by this behaviour. 

I might simply be conflating separate issues, but please free to test on your side. (E.g. Set a browser breakpoint [here](https://github.com/grantmcdermott/plot2/blob/0c7db77cf96e01cb969f8aea6c8768f91e16284a/R/plot2.R#L526) and run my second example above. If I do so and call `par("mfg")` immediately before and immediately after `par(omd = c(0, 1-w, 0, 1))` I get different results.)

zeileis:
Interesting, I wasn't aware of that either. If one of the outer `par()` settings is touched, it seems to reset the plot somehow. Other parameters can be altered. For example, this is ok:

```
par(mfrow = c(1, 2))
plot(1:5)
par(mar = c(8, 6, 1, 1))
plot(5:1)
```

But this re-initializes the cycling through the panels:

```
par(mfrow = c(1, 2))
[...]

PR #83, opened 2023-11-30
Add facet support

### LINKED ISSUE #79: Facet support

grantmcdermott:
Closes #79.

### Description

This PR introduces facet support to **plot2**. While user-facing changes are minimal&mdash;essentially all running through a new "facet" argument&mdash;it has required a fairly major refactoring of internal code to correctly interface with the  existing "by" grouping code logic. At a high level, this is what **plot2** does internally now:

- Grab the "by" and "facet" variables (if any).
- Open a new plot window, draw the legend (if "by" is not null) and add the other "common" plot elements like main title.
- Loop over the facets (if any) and set `par(mfrow)` to draw the axes, grid, frames and sub-titles of the individual facets. Do _not_ add any interior plot elements yet. We're just setting the facet windows for now.
- Finally, draw the interior plot elements via a nested loop:
   - Outer loop: Split the data by the "by" grouping variable and loop over each group sub-sample.
   - Inner loop: Split the group sub-sample by the "facet" variable and loop over each facet window (manually toggling `par(mfg)` to make sure that we do this in the correct order.)
   - Draw all the interior plot elements (points, ribbons, etc.) for the by-facet combination designated by this stage of the nested loop.
   - Repeat for the remaining iterations.

~~This internal code refactoring has broken a few tests. Some are just to do with plotting order&mdash;which elements are plotted on top of which&mdash;so not a serious blocker. A potentially more serious issue is that manually configuring a grid of plots via `par(mfrow)` no longer works consistently. I'm pretty sanguine about this and view it as an acceptable tradeoff (see also #65).~~ **UPDATE:** Fixed.

### Examples

``` r
#> ℹ Loading plot2

data("penguins", package = "palmerpenguins")
penguins = subset(penguins, !is.na(body_mass_g) & !is.na(species))

par(pch = 19)

# basic faceting on its own
with(
  penguins,
  plot2(
    x = flipper_length_mm, y = body_mass_g,
    facet = species
  )
)
```

``` r

# fancier version with additional args (incl. interaction with separate by arg)
with(
  penguins,
  plot2(
    x = flipper_length_mm, y = body_mass_g,
    by = sex, facet = species,
    palette = "dark2",
    grid = TRUE, frame = FALSE,
    main = "Palmer Penguins"
  )
)
```

``` r

# facet = "by" (redundant but there if you want it)
with(
  penguins,
  plot2(
    x = flipper_length_mm, y = body_mass_g,
    by = species, facet = "by",
    grid = TRUE, frame = FALSE,
    main = "Palmer Penguins"
  )
)
```

``` r

# density example
with(
  penguins,
  plot2(
    x = body_mass_g, facet = species,
    type = "density",
    fill = "black",
    grid = TRUE, frame = FALSE,
    main = "Penguin body mass (g)"
  )
)
```

``` r

# and another one (incl. interaction with a separate by arg)
with(
  penguins,
  plot2(
    x = body_mass_g, by = sex, facet = species,
    type = "density",
    fill = "by", palette = "dark2",
    grid = TRUE, frame = FALSE,
    main = "Penguin body mass (g)"
  )
)
```

<sup>Created on 2023-11-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### Chores for this PR

- [x] Right now, adding new elements to a faceted plot via `plot2(..., add = TRUE)` doesn't work perfectly because of some internal adjustments for facet titles. In other words the added elements can be a little off in terms of y-axis spacing. UPDATE: fixed with introduction of global par2 options. 
- [x] Add/update tests.
- [x] Add facet examples to help file.
- [x] Update vignette.

### Still to do / decide:

We don't necessarily have to resolve or roll these all into this PR. But here are some things that I think we ultimately need take a stance on w.r.t. facet support.

- [ ] Smarter arrangements of facets (e.g., "square" window arrangements) instead of current single row default.
- [ ] Smarter axes options (e.g., only print the axis ticks for the "outer" facets).
- [ ] Should we allow free axis scaling/limits for individual facets?
- [ ] Extra facets.args argument to control the above three options (maybe others)
- [ ] Decide whether to support (two-sided) formula argument, e.g.  `facet = ~facetvar` or `facet = facetrow  ~ facetcol` (for ggplot2-esque facet_grid).  
- [ ] If yes to the above, should we integrate formula support directly into the plot.formula method somehow... maybe via a second vertical bar, e.g. `y ~ x | by | facetrow ~ facetcol`?
- [ ] Should the larger legend cex be adjusted to match the smaller facet elements (or, vice versa)?


zeileis:
Grant, really cool, thanks for the work on this! I just want to comment on two aspects that we had visited before:

**Layout:** In my first attempt to add both faceting and categorical variables https://github.com/grantmcdermott/plot2/pull/12, I had just added an argument `mfrow` so that we could set `mfrow = TRUE` or `mfrow = c(1, 4)` etc. Maybe this would be feasible for letting the users decide whether they want square layouts or not.

**Axes:** It would be good to separate out the functions for drawing the panels and drawing the axes now. This might let us add the support for categorical variables from https://github.com/grantmcdermott/plot2/pull/12 that we had further discussed in https://github.com/grantmcdermott/plot2/issues/2. My suggestion was that there should be workhorse functions a la `plot2_xclass_yclass_plottype(x, y, by, axes = TRUE, ...)`.

From reading your description above (but not studying the code in detail), I thought that it might maybe help to have accompanying functions `axis2_xlass_yclass_plottype(...)` that could be inserted into the workflow for drawing the axes?

grantmcdermott:
Okay, I haven't managed to squeeze in everything---in particular, we'll need to address the single facet row soon---but I think that this PR is now ready to review and merge.

Here's one more example, demonstrating the add functionality works with facets.

``` r
pkgload::load_all("~/Documents/Projects/plot2")
#> ℹ Loading plot2

data("penguins", package = "palmerpenguins")
penguins = subset(penguins, !is.na(body_mass_g) & !is.na(species))

# Predict body mass with interactions to match facet-by combo further below
mod = lm(body_mass_g ~ flipper_length_mm * species * sex, data = penguins)

# Add predictions to dataset
penguins = cbind(penguins, predict(mod, newdata = penguins, interval = "confidence"))

# Original obs
with(
  penguins,
  plot2(
    x = flipper_length_mm, y = body_mass_g,
    by = sex, facet = species,
    palette = "dark2", fill = "by", 
    grid = TRUE, frame = FALSE,
    main = "Penguins"
  )
)

# Add model predictions
with(
  penguins,
  plot2(
    x = flipper_length_mm, y = fit, ymin = lwr, ymax = upr,
    by = sex, facet = species,
    type = "ribbon",
    palette = "dark2",
    add = TRUE
  )
)
```



<sup>Created on 2023-12-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

vincentarelbundock:
I can`t look at the code, but this looks absolutely amazing. Well done!

Personally, I would avoid putting too much burden on the formula. Faceting does feel like a different functionality which should mostly (exclusively?) be called with its own argument. Also, focusing on the `facet` argument eventually allow us to do more complex things like multiple variables with a two-sided formula indicating rows and columns.

The `mfrow` feature seems important and powerful. Is that name too generic? Will people think it does something else, like allow multiple `plot2()` calls? If it only applies to `facet`, maybe it should be `facet_mfrow`. I don't have a strong view...

Overall, very very nice stuff!

grantmcdermott:
Okay, let's merge this as-is then and tackle the remaining facet questions in standalone issue(s). I'll copy them across shortly.

PR #89, opened 2023-12-04
Setup GHA workflow for `altdoc`

etiennebacher:
This will render the docs in the branch `gh-pages` every time something is pushed to `main`. 

You can change the branch to which changes are pushed at the bottom of the yaml file, but I don't recommend that because pushing to `main` means that you and others will always need to update your local repo with the latest docs rendering. Using `gh-pages` will also allow to remove the `docs` folder from `main`.

This deployment works on my fork and takes about 2min in total.

Not sure you want that, I'm just proposing :)

### DIFF
--- a/.github/workflows/altdoc.yaml
+++ b/.github/workflows/altdoc.yaml
@@ -43,9 +43,8 @@ jobs:
 
       - name: Build site
         run: |
-          future::plan(future::multicore)
           install.packages(".", repos = NULL, type = "source")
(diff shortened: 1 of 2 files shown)


vincentarelbundock:
Yes, I would think so.

But you'll have to modify your repo settings to deploy the documentation from the root of  `gh-pages` branch instead of `docs/` of `main`.

etiennebacher:
Yes you can delete the `docs` folder on `main` (but not the `altdoc` one), and also modify the branch from which you deploy the website as Vincent said

PR #109, opened 2024-01-27
website: docsify -> quarto with gallery

vincentarelbundock:


vincentarelbundock:
I merged it immediately (sorry!) so we can see the live version. It's easy to roll back if you don't like it.

grantmcdermott:
Thanks!

PR #113, opened 2024-01-29
Website: add table of contents on each page

etiennebacher:
Just to make it easier to jump to examples for instance.

Before:

image

After:

image


grantmcdermott:
Dig it, thanks!

Issue #22, opened 2023-04-14
Package name: plot2 vs. baseplot vs. tinyplot vs. ?


zeileis:
I have been wondering whether `plot2()` is the best and/or most intuitive name for the package and its main function. It's certainly "ok" but I'm not excited about it. So I wanted to ask what you think, Grant @grantmcdermott and Vincent @vincentarelbundock, maybe others have thoughts/ideas as well.

A somewhat longer but more verbose alternative might be `baseplot()`. Short alternatives would be `draw()` or `viz()`. Currently, I think I slightly prefer `baseplot()` over `plot2()`. Other opinions or better alternatives?

vincentarelbundock:
Also not sure about `plot2`. To me, it suggests that it either wants to supersede `plot()`, or that its ultimate goal is to cover the feature set of `ggplot2`. Neither of which is quite the case.

I think I like `baseplot`; matches `basetheme`. `tinyplot` would also be an option, but I think I like that less.

I don't have great original ideas for package names, sorry. Will come back if I think of one.

grantmcdermott:
Thanks for kicking this discussion off @zeileis. I agree that this is something we need to resolve before a CRAN submission.

I'm not wedded to `plot2` and your pushback is certainly shifting my priors... But allow me to play the role of defence here for a moment. In my view, `plot2` has some key arguments in its favour:

- _Simplicity_. Not only only from the user side (just add a "2"), but also from the perspective of API management. If if we decide to add support for grouped histograms, for example, then `hist2()` offers itself as the natural and consistent function call.
- _Branding_. Funny enough, I actually think the implicit linkage to `ggplot2` is a feature, not a bug. I certainly don't envisage that we will support all, or even most, of what `ggplot2` does. But there is a non-trivial set of operations that I think we could cover that are (quote unquote) easy with `ggplot2`, but finicky with base plot. I would like `plot2` to serve as a natural alternative for this set of functions&mdash;grouped plots being the exemplar&mdash;and if some reflected <name>2 "brand recognition" helps steer people our way, all the better IMO.

I had one more point that escapes me for the moment and I have to run to pick up the kids. But will leave these two for the moment. For the record, I too prefer `baseplot` among the alternatives.

(Maybe we put it to popular vote on Tiwtter/Mastodon?)

zeileis:
Thanks, Grant, interesting! Just a couple of comments:

[...]

PR #114, opened 2024-01-30
update color reference from arXiv to The R Journal

zeileis:
Grant @grantmcdermott , just a mini-update in response to <https://fosstodon.org/@devSJR/111838661757048332>. Do you prefer if I send you such modifications as PRs or should I merge myself (or commit directly in main)?

### DIFF
--- a/vignettes/intro_tutorial.Rmd
+++ b/vignettes/intro_tutorial.Rmd
@@ -132,9 +132,9 @@ argument. The default group colours are inherited from the user's default
 palette. (Most likely the "R4" set of colors; see `?palette`). However, all of
 the various palettes listed by `palette.pals()` and `hcl.pals()` are supported
 as convenience strings.^[See the accompanying help pages of those two functions
-for more details on the available palettes, or read the
-[article](https://arxiv.org/pdf/2303.04918.pdf) by Achim Zeileis and Paul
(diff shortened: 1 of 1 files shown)


vincentarelbundock:
Grant can say what workflow he prefers, but this is trivial so I feel I can just review and merge now.

grantmcdermott:
Thanks, folks.

To Vincent's point. You should both definitely feel empowered to merge PRs, including your own, that you believe are simple and ready (i.e., pass all tests). For the bigger changes, I'd prefer us to have at least one other person review and I'd obviously like to be tagged in all the design-based changes.

(We can revisit a more formal process once the package hits CRAN. But I feel comfortable that we can just rely on individual judgment 'til then.)

zeileis:
Thanks for the explanation, Grant! That's perfect and works very well for me.
I definitely agree that bigger changes should be discussed first and wouldn't merge anything without tagging you first. It was really just about small typos and updates and stuff like that.

grantmcdermott:
Super.

Seeing as we're talking PRs... Just a HU that I'd like to resolve the package name issue (#22 ) as the priority for the next release, since that would obviously be a breaking change if we do proceed. I'll try to summarise my thoughts in the relevant thread when I get a sec and then we can look to make a final call by the end of this week or the next.

PR #128, opened 2024-02-17
skip tests on cran for smaller package

### LINKED ISSUE #111: Reduce size of installed package
This is > 5 MB, primarily due to all of the test snapshots in `inst/tinytest`.

As per R CMD check:

```
❯ checking installed package size ... NOTE
    installed size is  7.5Mb
    sub-directories of 1Mb or more:
      doc        3.0Mb
      help       1.5Mb
      tinytest   2.9Mb
```

Can we retain the GitHub testing infrastructure but skip it on CRAN? This is something that I don't have experience with, but I imagine there's a simple workaround.

vincentarelbundock:
Closes #111.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -27,6 +27,7 @@ jobs:
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       R_KEEP_PKG_SOURCE: yes
+      R_NOT_CRAN: yes
 
     steps:
       - uses: actions/checkout@v3
--- a/inst/tinytest/test-datetime.R
+++ b/inst/tinytest/test-datetime.R
@@ -1,5 +1,6 @@
 source("helpers.R")
 using("tinysnapshot")
+if (ON_CRAN) exit_file("CRAN")
 
 
 x_dtt = strptime(20010101:20010110, format="%Y%m%d", tz = "UTC")
--- a/inst/tinytest/helpers.R
+++ b/inst/tinytest/helpers.R
@@ -1,5 +1,7 @@
 library(tinytest)
 library(tinysnapshot)
+ON_CRAN <- !Sys.getenv("R_NOT_CRAN") %in% c("yes", "true", "TRUE")
+ON_LINUX <- Sys.info()["sysname"] == "Linux"
 options("tinysnapshot_os" = "Linux")
 options("tinysnapshot_device" = "svglite")
 options("tinysnapshot_device_args" = list(user_fonts = fontquiver::font_families("Liberation")))
--- a/inst/tinytest/test-README.R
+++ b/inst/tinytest/test-README.R
@@ -1,5 +1,6 @@
 source("helpers.R")
 using("tinysnapshot")
+if (ON_CRAN) exit_file("CRAN")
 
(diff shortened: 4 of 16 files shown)


grantmcdermott:
Thanks for doing this! 

Over the weekend, I had the idea of hosting the test snapshots and infra in a git submodule. I _think_ this could work in a way that (a) gets you full CI testing and (b) avoids the manual Rbuildignore step. But I have to experiment with a dummy package first. I'll pause on merging this until I've had a chance to try it out.

.github/workflows/R-CMD-check.yaml
```diff
@@ -27,6 +27,7 @@ jobs:
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       R_KEEP_PKG_SOURCE: yes
+      R_NOT_CRAN: yes
```
eitsupi, line 30: Is there any reason not to use the widely used `NOT_CRAN="true"`?

vincentarelbundock: no good reason, no

eitsupi: I recommend using `NOT_CRAN`.

For example, the r-lib's action has already set it as shown below.
https://github.com/grantmcdermott/tinyplot/actions/runs/8009321355/job/21877708151?pr=122#step:6:16

inst/tinytest/test-datetime.R
```diff
@@ -1,5 +1,6 @@
 source("helpers.R")
 using("tinysnapshot")
+if (ON_CRAN) exit_file("CRAN")
```
grantmcdermott, line 4: I might just be missing something obvious, but is there a reason the ON_LINUX  flags aren't included in some of these files?

vincentarelbundock: I guess they could be, but you don't have any other OS in the github actions matrix, and you probably only run the tests on a linux machine, so it feels superfluous.

.github/workflows/R-CMD-check.yaml
eddelbuettel: At GitHub you can simple check for `CI` or `GITHUB_ACTION` being present.  Actually checking for something that has to be present for an action to take place always strikes me as cleaner.  But that may just be me.

grantmcdermott: Okay, let's go for `NOT_CRAN` per @eitsupi's comment, since that seems to be a norm.

inst/tinytest/test-datetime.R
grantmcdermott: I know it cuts against the tinytest ethos, but I wonder whether she should just add these flags to the `helpers.R` file, so that they are automatically enabled for all test scripts (instead of adding individually to each one). Thoughts?

vincentarelbundock: Sure. Or include a similar logic in `/tests/tinytest.R` do skip the whole run.

grantmcdermott:
Sorry, I seem to have done something daft in committing extra changes to @vincentarelbundock's fork. I've opened up a separate PR here that integrates everything (including @eitsupi's suggestions) here: https://github.com/grantmcdermott/tinyplot/pull/145

Closing this one to avoid duplication.

PR #159, opened 2024-07-06
Support type = "polypath"

### LINKED ISSUE #158: `"polypath"`

grantmcdermott:
Closes #158.


grantmcdermott:
Aside: This should also enable consistent **tinyplot** support for spatial objects, since `polypath` is what **sf**, **geos**, and co. use under the hood for their `plot` methods.


``` r
#> ℹ Loading tinyplot
```

``` r
library(sf)
#> Linking to GEOS 3.12.2, GDAL 3.9.0, PROJ 9.4.1; sf_use_s2() is TRUE
```

``` r

nc_shapefile = system.file("shape/nc.shp", package = "sf")
nc = st_read(nc_shapefile)
#> Reading layer `nc' from data source `/usr/lib/R/library/sf/shape/nc.shp' using driver `ESRI Shapefile'
#> Simple feature collection with 100 features and 14 fields
#> Geometry type: MULTIPOLYGON
#> Dimension:     XY
#> Bounding box:  xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> Geodetic CRS:  NAD27
```

``` r

get_geom = function(x) {
    m = Map(as.data.frame, Map(`[[`, Map(`[[`, nc$geometry, 1), 1))
    for (i in seq_along(m)) m[[i]][["idx"]] = i
    m = do.call("rbind", m)
    names(m) = c("lon", "lat", "idx")
    m
}

ncg = get_geom(nc)
plt(
    lat ~ lon | factor(idx), ncg,
    type = "polypath",
    fill = 0.8, col = "white",
    legend = FALSE
)
```



<sup>Created on 2024-07-06 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

edzer:
I think flattening polygon geometry into a data.frame is what `ggplot2` tried to do with `fortify` on `sp` objects (i.e., before `geom_sf()`). It worked for simple cases, but not for all, IIRC because of holes in polygons.

PR #164, opened 2024-07-11
Support type = "histogram" (alias "hist")

### LINKED ISSUE #21: Wishlist: Refined methods for densities/histograms
In addition to the `plot2.density` method from PR https://github.com/grantmcdermott/plot2/pull/18, it would be nice to have a formula-based interface:

    plot2(~ x | by, type = c("density", "histogram"), ...)

where it can be selected which display is used for `x` in each `by` group.

In addition to just lines it would be nice to support semi-transparent shaded areas below the lines as well (via `polygon()`). The area might also be opaque if we order the by groups (e.g., by median) from highest to lowest, which would yield a kind of ridge lines plot.

_GM edit: adding task items to track._

- [x] #61
- [x] #56
- [ ] #72

### LINKED ISSUE #72: Support type = "histogram"

grantmcdermott:
Closes #21. Closes #72.

## Examples

I use the one-sided formula method for the below examples. But the atomic method (e.g., ` plt(Nile, type = "histogram")`) works just the same.

``` r
#> ℹ Loading tinyplot
```

### Standard histogram

``` r
plt(~Petal.Length, iris, type = "hist") # "hist" is alias for "histogram"
```

**Question:** I haven't switched off the surrounding plot frame (box) by default. Should we do this to make it more consistent with base `hist()`?

``` r
plt(~Petal.Length, iris, type = "hist", frame = FALSE)
```

### Grouped histograms

``` r
plt(~Petal.Length | Species, iris, type = "hist")
```

**Question:** By default, the fill of grouped histograms is solid. Should we rather enable automatic alpha fill transparency for these (see next examples)?

Aside: We can adjust the number of breaks by passing a `breaks` argument.

``` r
plt(~Petal.Length | Species, iris, type = "hist", fill = 0.5, breaks = 30)
```

### Grouped and faceted histograms

We could also facet by itself (i.e., no colour grouping), although I don't show that here.

``` r
plt(~Petal.Length | Species, iris, type = "hist", fill = 0.5, breaks = 30, facet = "by")
```

Again, we haven't turned off the frame by default but we might want to...

``` r
plt(~Petal.Length | Species, iris, type = "hist", fill = 0.5, breaks = 30, facet = "by", frame = FALSE)
```

More elaborate example where we use different (colour) groups and facets.

``` r
iris2 = iris |>
  transform(long_sepal = Sepal.Length > mean(Sepal.Length))
plt(~Petal.Length | Species, iris2, type = "hist", fill = 0.5, breaks = 30, facet = ~long_sepal)
```

<sup>Created on 2024-07-11 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


grantmcdermott:
@zeileis I know that you're busy at userR!2024, but do you want to kick the tyres on this one given your original issue in #21?

zeileis:
Sorry, Grant, too much input at useR! to process within one week... If you want to move ahead, you can of course do so. Otherwise I'll try to take a closer look this week!

zeileis:
First of all, very nice and useful.

Regarding the box/frame: I agree with you and would keep it. I'm not sure whether there is a particularly important reason for `hist()` not to draw it. But both lattice and and ggplot2 (bw) have it. So I'd rather be consistent with the other tinyplot types than with the corresponding standalone base plot. In any case, it's easy to switch off.

Regarding the alpha transparency I think that we need some by default. In the multigroup case, the groups will typically overlap and then the display without transparency can be very hard to read. Also, the default of `hist()` - black border lines with light gray fill - could also be interpreted as opaque border with transparent fill. So we could adopt that strategy to cover both the single group and multigroup case.

Also, I would expect the histogram to look similar to a filled density:

    tinyplot(~ Sepal.Length | Species, data = iris, type = "density", palette = "Dark3", fill = "by")

tinyplot-density

So maybe more like:

   tinyplot(~ Sepal.Length | Species, data = iris, type = "hist", palette = "Dark3", breaks = 20, alpha = 1/3)

tinyplot-hist-alpha

Rather than:

   tinyplot(~ Sepal.Length | Species, data = iris, type = "hist", palette = "Dark3", breaks = 20)

tinyplot-hist

Final comment: I'm not sure whether I would mimic the default main title of `hist()` but rather be consistent across `tinyplot()` types.

grantmcdermott:
Suggested changes have been actioned, thanks @zeileis!

vincentarelbundock:
these look fantastic, BTW

PR #165, opened 2024-07-15
Fix some lints

etiennebacher:
Hi @grantmcdermott, I've made a new package to find lints and automatically fix them ([`flint`](https://flint.etiennebacher.com/)). I used `tinyplot` as one of my test cases and found a few lints so I thought you might want to remove them.

FYI those fixes were automatically generated with `flint::fix(exclude_linters = "equal_assignment")`. End of the self-advertising :)

### DIFF
--- a/R/tinyplot.R
+++ b/R/tinyplot.R
@@ -1656,13 +1656,13 @@ tinyplot.formula = function(
     # special catch if by is the same as x or y (normally for continuous legend)
     by_same_y = by_same_x = FALSE
     fml_all_vars = all.vars(m$formula, unique = FALSE)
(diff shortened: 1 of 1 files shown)


grantmcdermott:
Ah, super thanks @etiennebacher ... and very cool new package. Are you working towards a CRAN release, or are you going to run up against a Rust version requirement (_a la_ polars)?

etiennebacher:
Yes the plan is to eventually have it on CRAN. `flint` shouldn't be problematic, but it is built on my other package `astgrepr` that calls Rust code, so if there's any issue it's from this one. But I will try to put both on CRAN during the summer.

PR #171, opened 2024-07-21
refactor: setup_device()

vincentarelbundock:
The `tinyplot()` function is currently over 1100 lines long, and I think it would be nice to add new features.

Would you be open to a few small PRs like this one, to split up the function in smaller, more manageable blocks?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -49,6 +49,7 @@ Suggests:
   altdoc (>= 0.3.0),
   basetheme,
   fontquiver,
+  png,
   rsvg,
   svglite,
   tinytest,
--- a/R/setup_device.R
+++ b/R/setup_device.R
@@ -1,0 +1,45 @@
+setup_device = function(file, width, height) {
+  # write to file
+  if (!is.null(file)) {
+    filepath = file
+    filewidth = width
+    fileheight = height
+    if (is.null(filewidth)) filewidth = .tpar[["file.width"]]
+    if (is.null(fileheight)) fileheight = .tpar[["file.height"]]
+    fileres = .tpar[["file.res"]]
+    # catch to close interactive device if one isn't already open
+    fkdev = is.null(dev.list())
+    # grab existing device pars to pass on to next one
+    dop = par(no.readonly = TRUE)
+    # close interactive device if not already open
+    if (isTRUE(fkdev)) dev.off()
+    exttype = file_ext(filepath)
+    if (exttype == "jpg") exttype = "jpeg"
+    switch(exttype,
+      png = png(filepath, width = filewidth, height = fileheight, units = "in", res = fileres),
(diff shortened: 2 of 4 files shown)


grantmcdermott:
Yeah, it has gotten unwieldy. I'm definitely open to PRs like this, but I will say that I've found the plot system is sensitive to environmental scoping (since everything is drawn immediately). Reworking `draw-legend.R` into separate function was quite a pain, for example, and I had to introduce several auxiliary flags to make sure things were being passed correctly in and out.

So caveat emptor. But again, please free to go for it.

vincentarelbundock:
Yeah, I can image this being fiddly.

Do you have some impressionistic sense of how much I can trust rest. For example, with the PR I tried 3-4 interactive graphs, 3-4 save to files, and the test suite passes. Is that enough?

grantmcdermott:
> Do you have some impressionistic sense of how much I can trust the test suite. For example, with this PR I tried 3-4 interactive graphs, 3-4 save to files, and the test suite passes. Is that enough?

Test coverage is pretty thorough, albeit only implemented for Linux (as you probably recall).

In fact, the only place I'd say there are gaps is with the file writing/saving logic, i.e. partly what you're adjusting here :-/ 
I just wasn't sure how to effectively incorporate this into the `tinysnapshot` workflow.

Maybe we could try something simple like the following... We'd need to include {png} in **Suggests** and then add the following test to  `inst/tinytest/test-misc.R`:

```r
# test file saving and dimensions
if (requireNamespace("png", quitely = TRUE)) {
  f = function () {
    tmp_path = tempfile(fileext = ".png")
    tinyplot(
      Sepal.Length ~ Petal.Length, data = iris,
      file = tmp_path, width = 4, height = 4
    )
    obj = png::readPNG(tmp_path, info = TRUE)
    unlink(tmp_path)
    dims = attr(obj, "dim")
    return(dims)
  }
  expect_equal(f(), c(1200, 1200, 4), label = "png_size")
}
```

Do you mind incorporating this change as part of your PR? I'm happy to merge, assuming this new test passes and given your manual checks too.

grantmcdermott:
Thanks!

zeileis:
Thanks to both of you for the all the work and active discussions. I just returned after being away with the family for the weekend: Are there any issues/PRs I can still help with?

Modularizing the building blocks of `tinyplot()` would indeed be great. Then I would like to contribute a `type = "spine"` as I mentioned a couple of times. Regarding the problems with environments and passing around certain arguments or attributes: We could set up a dedicated environment within the package (say `tinyplot_env`) and

* Empty the environment at the beginning (possibly unless `add = TRUE`).
* Copy data and other variables there.
* Make all modifications there.
* Use all the prepared information for drawing at the end.

Maybe this could even be taken to the point where we can store a copy of the environment and re-plot it later. But I haven't thought this through.

PR #183, opened 2024-08-02
Modularize more

vincentarelbundock:
This is an attempt to further split up the core function into helpers.

Overall, it seems to work fine, but the background (dotted) grid is misaligned in one snapshot. I must be passing an argument wrong.

@grantmcdermott do you know what variable is in charge of offsetting the grid so it aligns properly? Knowing that would help me diagnose.

restore_par_bottom


grantmcdermott:
The panel grid logic is specified here:
https://github.com/grantmcdermott/tinyplot/blob/e1293723808d245873210287a3dd82113933bbd1/R/tinyplot.R#L1192-L1217

Note the part about the custom `abline(... = pretty(extendrange(...)), ...)` logic instead of `grid()` to better handle generic classes like dates. So I think the answer to your question is `pretty(extendrange(...)`.

What happens if you redo the test but switch out `grid = grid()` for `grid = TRUE`? I assume it will pass, although the fact that `grid()` doesn't work here does me uncomfortable, since some knowledge of the plot construction is not being being passed through correctly.

vincentarelbundock:
Thanks for the pointer. That was helpful.

I found the issue and think this is ready for review.

There is no new code here. All I did was create two helper functions to extricate more complicated things out of `tinyplot.R`.

Unfortunately, I think this will conflict with https://github.com/grantmcdermott/tinyplot/pull/190  The relevant code block has not changed, but it was moved to a new helper function and file. You'll need to decide what you want to merge first.

grantmcdermott:
Excellent. I think it will be simpler to merge #190 first and then we can come back to this afterwards.

Issue #182, opened 2024-08-02
Possible to remove also the axes when frame=FALSE and grid=TRUE?


pglpm:
Thank you for this fantastic lightweight plot function!

It's very close to something I wrote for my own use, so I'd be happy to switch to tinyplot, which is more carefully developed and maintained :)

I was wondering if there's an option to eliminate the thick axes bars completely when one has `frame = FALSE` and `grid = TRUE`, as in the enclosed screenshot.

If not, could this be added to the wishlist? :)

![Screenshot_20240802_123941](https://github.com/user-attachments/assets/f8c5d63b-51d5-4b02-8170-1519b85f491c)

grantmcdermott:
Thanks for the kind words @pglpm.

RE: your FR. It's definitely something I'd like to support (and have wanted to for a while). What need to do is settle on a consistent argument interface. I think `axes = FALSE` needs to retain its current behaviour of removing all axes elements to ensure consistency with base `plot`. But perhaps we could support switching off the lines only with something like `axes = NA`?

P.S. A broader goal for the next release of `tinyplot` is to support theming elements. So a user could just call, say, `tinytheme("minimal")` and it would invoke this setting for all plots.

zeileis:
If `axes` can take values other than `TRUE` and `FALSE` in `tinyplot()`, one could also go for something like `axes = "ticks"` (or another verbal description). This is easier to extend in case further types should be added in the future.

Instead of using `axes` one could also use `xaxt` and `yaxt` which are intended to be type specifications via character descriptions. The problem is just that de facto only `"s"` (standard) and `"n"` (none) are supported.

grantmcdermott:
Good points @zeileis. We actually don't support `x(y)axt` in `tinyplot()`, so that's something that should probably be fixed too.

So, is your thinking that we should support a `"t"` ("or `"ticks") argument option for `axes`, `xaxt` and `yaxt`?

zeileis:
My idea was to just go through `axes` and to come up with new descriptive names (such as `"ticks"`) rather than just abbreviations (like `"t"`). The latter could be used by the end-users nevertheless due to `match.arg()`.

However, compatibility with `xaxt`/`yaxt` might not be a bad idea. The catch is, though, that `xaxt = "t"` currently already works but has a different meaning. More precisely, the supported specifications are currently:

* `"s"`: standard (default)
* `"n"`: none
[...]

PR #190, opened 2024-08-03
Fine control of axis types in axes/xaxt/yaxt

zeileis:
As discussed in https://github.com/grantmcdermott/tinyplot/issues/182 I have tried to implement character specifications of different axis types. The docs of the relevant arguments in `tinyplot.default` now say:

  * `axes`: logical or character. Should axes be drawn (`TRUE` or `FALSE`)? Or alternatively what type of axes should be drawn: `"standard"` (with axis, ticks, and labels; equivalent to `TRUE`), `"none"` (no axes; equivalent to `FALSE`), `"ticks"` (only ticks and labels without axis line), `"labels"` (only labels without ticks and axis line), `"axis"` (only axis line and labels but no ticks). To control this separately for the two axes, use the character specifications for `xaxt` and/or `yaxt`.
* `xaxt`, `yaxt`: character specifying the type of x-axis and y-axis, respectively. See `axes` for the possible values.

Illustration:

```
library("tinyplot")
x <- 0:100/10
y <- sin(x)
tinyplot(y ~ x, type = "l", lwd = 2, col = "steelblue", grid = TRUE, main = 'default')
tinyplot(y ~ x, type = "l", lwd = 2, col = "steelblue", grid = TRUE, main = 'axes = "ticks"', axes = "ticks")
tinyplot(y ~ x, type = "l", lwd = 2, col = "steelblue", grid = TRUE, main = 'axes = "labels"', axes = "labels")
tinyplot(y ~ x, type = "l", lwd = 2, col = "steelblue", grid = TRUE, main = 'axes = "axis"', axes = "axis")
tinyplot(y ~ x, type = "l", lwd = 2, col = "steelblue", grid = TRUE, main = 'xaxt = "ticks", yaxt = "labels"', xaxt = "ticks", yaxt = "labels")
tinyplot(y ~ x, type = "l", lwd = 2, col = "steelblue", grid = TRUE, main = 'axes = "none", yaxt = "standard"', axes = "none", yaxt = "standard")
```

Different tinyplot axes definitions

Note that the plot in the bottom left panel is maybe unexpected. The reason is that `frame.plot` is solely selected based on `axes` (which is `TRUE` here) and not based on `xaxt`/`yaxt`. One could consider changing that but I felt that the selection of the defaults is already involved enough for a first proposal.

Feel free to make any changes or discard the PR if it's not useful (in this way).

### DIFF
--- a/R/tinyplot.R
+++ b/R/tinyplot.R
@@ -112,9 +112,13 @@
 #' @param ylab a label for the y axis, defaults to a description of y.
 #' @param ann a logical value indicating whether the default annotation (title
 #'   and x and y axis labels) should appear on the plot.
-#' @param axes a logical value indicating whether both axes should be drawn on
-#'   the plot. Use `graphical parameter` "xaxt" or "yaxt" to suppress just one of
-#'   the axes.
+#' @param axes logical or character. Should axes be drawn (`TRUE` or `FALSE`)?
+#'   Or alternatively what type of axes should be drawn: `"standard" (with
+#'   axis, ticks, and labels; equivalent to `TRUE`), "none" (no axes;
+#'   equivalent to `FALSE`), `"ticks"` (only ticks and labels without axis line),
+#'   `"labels"` (only labels without ticks and axis line), `"axis"` (only axis
+#'   line and labels but no ticks). To control this separately for the two
+#'   axes, use the character specifications for `xaxt` and/or `yaxt`.
 #' @param frame.plot a logical indicating whether a box should be drawn around
 #'   the plot. Can also use `frame` as an acceptable argument alias.
 #' @param grid argument for plotting a background panel grid, one of either:
@@ -267,6 +271,8 @@
 #' @param height numeric giving the plot height in inches. Same considerations as
 #'  `width` (above) apply, e.g. will default to `tpar("file.height")` if not
 #'  specified.
+#' @param xaxt,yaxt character specifying the type of x-axis and y-axis, respectively.
+#'   See `axes` for the possible values.
 #' @param ... other graphical parameters. See \code{\link[graphics]{par}} or
 #'   the "Details" section of \code{\link[graphics]{plot}}.
 #'
@@ -506,6 +512,8 @@ tinyplot.default = function(
     width = NULL,
     height = NULL,
     empty = FALSE,
+    xaxt = NULL,
+    yaxt = NULL,
     ...
     ) {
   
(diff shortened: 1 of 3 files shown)


zeileis:
Unrelated bonus question which I noticed when preparing the example enough. Is it intentional that for the `x` and `y` sinus example from above, the following plots differ? The former is just black while the latter gets a continuous palette.

```
tinyplot(y ~ x)
tinyplot(sin(x) ~ x)
```

grantmcdermott:
This looks great, thanks @zeileis. I'm out with the family atm but will be and to review properly when I'm back home.

zeileis:
While mowing the lawn I thought that it would be better to handle all defaults of `axes`, `xaxt`, `yaxt`, and `frame.plot` early on in `tinyplot.default`. This means that the could is in a different place than the actual drawing of the axes but the defaults could be handled in a cleaner and more intelligible way. The default of `frame.plot` is now `NULL` and then we can do: https://github.com/grantmcdermott/tinyplot/blob/axis-type/R/tinyplot.R#L532-L545

This still produces the same plots as above except for the bottom left panel which is now drawn without a frame.

R/tinyplot.R
```diff
@@ -519,6 +529,32 @@ tinyplot.default = function(
...
+  axis_types = c("standard", "none", "labels", "ticks", "axis")
+  axes = match.arg(axes, axis_types)
+  if (is.null(xaxt)) xaxt = axes
+  if (is.null(yaxt)) yaxt = axes
+  xaxt = substr(match.arg(xaxt, axis_types), 1L, 1L)
+  yaxt = substr(match.arg(yaxt, axis_types), 1L, 1L)
+  axes = any(c(xaxt, yaxt) != "n")
+  if (is.null(frame.plot) || !is.logical(frame.plot)) frame.plot = all(c(xaxt, yaxt) %in% c("s", "a"))
+
+  ## auxiliary function with axis type
```
grantmcdermott, line 547: My vote is to move to its own file as an auxiliary function.

grantmcdermott (changes requested):
@zeileis This is great. Your plot examples look very slick!

My only minor request is to make `tinyAxis()` an unexported function (as you suggested).

R/tinyplot.R
zeileis: Sure, will do. Any name preferences?

- [ ] `tinyAxis()`
- [ ] `tiny_Axis()`
- [ ] `tiny_axis()`
- [ ] Other

grantmcdermott: I like `tinyAxis` for consistency.

zeileis: :+1:

zeileis:
Just in case `tinyAxis()` also gets called elsewhere in the future, I improved the implementation and made sure that the arguments enforced through the `type` do not conflict with the `...` arguments.

grantmcdermott:
Closes #182 .

grantmcdermott (approved):

grantmcdermott:
Thanks @zeileis!

PR #196, opened 2024-08-05
Boxplot widths

### LINKED ISSUE #195: Boxplots widths too narrow if x == by
``` r
library(tinyplot)

tinyplot(
  Petal.Length ~ Species, iris,
  type = "boxplot"
)
```

``` r
tinyplot(
  Petal.Length ~ Species | Species, iris,
  type = "boxplot",
  legend = list("topleft", bty = "o")
)
```

The bottom boxplots should be the same width as the top ones. But they're too narrow.

I think I know why this is happening (related to a factor coercion) and should hopefully be a simple fix.

grantmcdermott:
Closes #195.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@ Bug fixes:
 
 - Fix duplicate plots produced with `type = "density"`, which was a regression
 accidentally introduced in v0.2.0 (#187 @grantmcdermott)
+- Ensure correct boxplot positioning if `x` == `by`, or these two are
+functionally identical. (#196 @grantmcdermott)
 
 New Features:
 
@@ -25,6 +27,11 @@ the `xaxt` and `yaxt` for _separately_ controlling the two axes using the same
 keyword options. For example, `plt(0:10, xaxt = "l", yaxt = "t")` will yield a
 plot where the x-axis only contains labels and the y-axis contains both labels
 and ticks, but no axis line. (#190 @zeileis)
+- Support additional boxplot arguments like `varwidth`, `notch`, etc. Note
+that `tinyplot(..., type = "boxplot", boxwidth = <num>)` is equivalent to the
+`boxplot(..., width = <num>)`; we just use the "box(width)" prefix to avoid 
+conflicting with the existing `tinyplot(..., width)` argument.
+(#196 @grantmcdermott)
 
 Internals:
 
--- a/R/draw_elements.R
+++ b/R/draw_elements.R
@@ -111,12 +111,13 @@ draw_elements = function(
           rule = irule
         )
       } else if (type == "boxplot") {
+        at_xx = unique(xx)
+        x_by = !is.null(by) && identical(xlvls, lgnd_labs) 
         horizontal = ifelse(!is.null(dots[["horizontal"]]), dots[["horizontal"]], FALSE)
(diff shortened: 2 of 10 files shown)


grantmcdermott:
@zeileis I think that this should address the various examples that we discussed in #195. Let me know if you have time to do a quick review / kick the tyres. Thanks!

zeileis:
Thanks, will have a look tonight!

zeileis:
The boxplots look great, thanks Grant @grantmcdermott ! I played around with various combinations of `by` and `facet` and `boxplot`-specific arguments. Everything worked as intended except one extremely contrived example (see below).

I mainly wondered about the computation of `x_by`. First, why is it set to `FALSE` except for `type == "boxplot"` in

https://github.com/grantmcdermott/tinyplot/blob/boxplot_widths/R/tinyplot.R#L654-L658

Second, it is done rather early now but `x` and `by` might have been modified already in `tinyplot.default`.

As for the first point: I don't understand the logic why this should be limited to boxplots. The logic should also apply to other plot types, even if these choose to ignore the `x_by` argument.

And an example, where the `x_by` gets `FALSE` even though it should be `TRUE` is if both variables are numeric. The reason is that `by` gets coerced to factor before the comparison. Due to this the following plots are not the same, although they should:

```
tinyplot(mpg ~ factor(gear) | factor(gear), facet = "by", data = mtcars, type = "boxplot")
tinyplot(mpg ~ gear | gear, facet = "by", data = mtcars, type = "boxplot")
```

grantmcdermott:
Ah, good catch! 

Let me take your points in reverse order:

> Due to this the following plots are not the same, although they should:

Thanks. I bumped this `x_by` assignment further up in the code (per your helpful suggestion) to avoid any earlier coercion "mismatches". So the examples that you gave should now be working correctly. I included a test along these lines.

> I mainly wondered about the computation of x_by. First, why is it set to FALSE except for type == "boxplot" in

The `x_by` flag only matters (i.e., is only used) for boxplots. So I was trying to avoid doing unnecessary evaluations for other plot types, which might slow down the overall `tinyplot()` function. I admit that this probably doesn't make a material difference, since `identical` is normally fast. But I do like the principle of only doing work you need  and would really like to preserve the overall snappiness of `tinyplot`. (I don't want to be dogmatic about this though, and could be convinced to evaluate for other plot types if you think it best.)

zeileis:
Thanks for the update! Regarding the second point: I agree that avoiding unnecessary computations is a good idea in general - but so is modularity. The more we have these scattered special cases in `tinyplot.default` the more difficult it becomes to make the plot types self-contained.

Another option (I haven't tried it) could be to compare something like `identical(as.integer(x), as.integer(by))` or something along those line **within** the boxplot code (rather than doing it before).

grantmcdermott:
> I agree that avoiding unnecessary computations is a good idea in general - but so is modularity. The more we have these scattered special cases in tinyplot.default the more difficult it becomes to make the plot types self-contained.

Okay, done. Feel free to squash+merge if you are happy.

zeileis:
Great, thanks, Grant! Should I also delete the branch after merging it?

grantmcdermott:
Thanks! (Free free to delete the branch in the future. I'll do it here.)

Issue #191, opened 2024-08-03
`tinyplot(fn(x) ~ x)` triggers inadvertent continuous groups


grantmcdermott:
First spotted by @zeileis here: https://github.com/grantmcdermott/tinyplot/pull/190#issuecomment-2267002593

Examples:

``` r
library(tinyplot)
x = 0:100/10

tinyplot(sin(x) ~ x)
```

![](https://i.imgur.com/GPBQrMU.png)

``` r
tinyplot(log(x+1) ~ x)
```

![](https://i.imgur.com/cPwCRtC.png)

``` r
tinyplot(jitter(x) ~ x)
```

![](https://i.imgur.com/X88v3ER.png)

Note that this only appears to affect the formula method. The equivalent atomic-methods plots do not produce the continuous groups:

``` r
tinyplot(x, sin(x))
```

![](https://i.imgur.com/V846Xkx.png)

``` r
tinyplot(x, log(x+1))
```

![](https://i.imgur.com/8yOqY8C.png)

``` r
tinyplot(x, jitter(x))
```

![](https://i.imgur.com/4elixhZ.png)

Similarly, it's only the case for `fn(x) ~ x`. Running `fn(y) ~ x` doesn't trigger this behaviour.

zeileis:
Maybe we should use `format()` instead of `paste()` in `paste(attr(terms(m$formula), "variables")[[2]])` in https://github.com/grantmcdermott/tinyplot/blob/main/R/tinyplot.R#L1455

zeileis:
Hmm, additionally, `all.vars()` does not preserve the transformation labels...will have to dig deeper.

zeileis:
I didn't find a base R function that can extract the `term.labels` with both (a) transformations and (b) duplicates. The `term.labels` from the `terms()` preserve (a) but not (b). And `all.vars()` can preserve (b) but not (a).

Assuming that all the terms in our formulas are separated by `+` (which we used to replace `|`), the following might work:

```
all.terms = function(formula) {
  if (inherits(formula, "formula")) {
    if (length(formula) == 3L) {
      rval = deparse(formula[[2L]])
      formula = formula[[3L]]
    } else {
      rval = character(0L)
      formula = formula[[2L]]
    }
    rval <- c(rval, all.terms(formula))
  } else if (inherits(formula, "call") && formula[[1L]] == as.name("+")) {
    rval <- c(all.terms(formula[[2L]]), all.terms(formula[[3L]]))
[...]

PR #197, opened 2024-08-06
Formula processing

zeileis:
Fixes https://github.com/grantmcdermott/tinyplot/issues/191

I started to work on a completely revised formula handling to address the issue above. This is still leaner than employing the `Formula` package but uses a similar approach:

- Auxiliary function `tinyformula` explicitly breaks up the formulas into separate formulas for `x`, `y` (if any), `by` (if any), and `facet` (if a formula) plus a `full` formula combining all elements (on the right-hand side).
- Auxiliary function `tinyframe` to extract the variables from the different formulas from the overall model frame.
- This allows for better sanity checking, e.g., that there is exactly one `x` and at most one `y` etc.

By and large things seem to work. However, I have apparently messed up the two-sided `facet` formulas. Grant @grantmcdermott if you have the time to look at this, this would be great. I'm not sure whether I fully understand how the attributes work...

### DIFF
--- a/R/facet.R
+++ b/R/facet.R
@@ -7,14 +7,9 @@ facet_layout = function(facet, add = FALSE, facet.args = list()) {
   nfacet_rows = 1
   nfacet_cols = 1
   if (!is.null(facet)) {
-    if (is.factor(facet)) {
-      facets = levels(facet)
-    } else {
-      facets = sort(unique(facet))
-    }
+    facets = if (is.factor(facet)) levels(facet) else sort(unique(facet))
     ifacet = seq_along(facets)
     nfacets = length(facets)
-    
     if (isTRUE(add)) {
       omfrow = par("mfrow")
       nfacet_rows = omfrow[1]
--- a/R/facet.R
+++ b/R/facet.R
@@ -16,6 +16,9 @@ facet_layout = function(facet, add = FALSE, facet.args = list()) {
       nfacet_rows = omfrow[1]
       nfacet_cols = omfrow[2]
     } else {
+      if (isTRUE(attr(facet, "facet_grid"))) {
+        facet.args[["nrow"]] = attr(facet, "facet_nrow")
+      }
       if (!is.null(facet.args[["nrow"]])) {
         nfacet_rows = facet.args[["nrow"]]
         nfacet_cols = ceiling(nfacets / nfacet_rows)
--- a/R/facet.R
+++ b/R/facet.R
@@ -7,7 +7,11 @@ facet_layout = function(facet, add = FALSE, facet.args = list()) {
   nfacet_rows = 1
   nfacet_cols = 1
   if (!is.null(facet)) {
-    facets = sort(unique(facet))
+    if (is.factor(facet)) {
+      facets = levels(facet)
+    } else {
+      facets = sort(unique(facet))
+    }
     ifacet = seq_along(facets)
     nfacets = length(facets)
     
--- a/R/tinyformula.R
+++ b/R/tinyformula.R
@@ -24,20 +24,20 @@ tinyformula = function(formula, facet = NULL) {
   facet = if (is.null(orig_facet) || !inherits(orig_facet, "formula")) NULL else ~ a + b
 
   ## fill with actual terms
-  environment(x) = parent.frame()
+  environment(x) = environment(formula)
   if (!is.null(y)) {
-    environment(y) = parent.frame()
+    environment(y) = environment(formula)
     y[[2L]] = formula[[2L]]
   }
   if (is.null(by)) {
     x[[2L]] = formula[[nf]]
   } else {
-    environment(by) = parent.frame()
+    environment(by) = environment(formula)
     by[[2L]] = formula[[nf]][[3L]]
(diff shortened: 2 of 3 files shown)


grantmcdermott:
Thanks so much for taking this on @zeileis!

> By and large things seem to work. However, I have apparently messed up the two-sided facet formulas. Grant @grantmcdermott if you have the time to look at this, this would be great. I'm not sure whether I fully understand how the attributes work...

This week is tough for me due to work deadlines, but I'll try to take a deeper look over the weekend.

zeileis:
That's perfectly fine. I'll be on vacation for a bit more than two weeks. So I won't have a lot of time either but should be able to respond to smaller messages.

I hope that the new code is also reasonably easy to read and understand. Most of it works fine but I still didn't understand how the `facet = a ~ b` specification is processed correctly. Some cases I was able to handle correctly but others I didn't understand properly. BTW, in principle we would now also have the infrastructure to have `facet` specifications with multiple variables on both siders, e.g., `a1 + a2 ~ b1 + b2` etc. But I didn't actually test this, yet, because I couldn't get even the simple case right.

grantmcdermott:
Hi @zeileis. I finally had some time to review this and everything looks excellent. Thank you!

I only have two minor comments and then we can merge:

- I didn't quite understand your comment about having "messed up two-sided formulas". Everything seems to be working as expected on my side and all the tests are passing. Is this comment outdated?
- Do you mind editing your opening comment above, by adding `"Fixes #191"` as a standalone line? This should ensure that the original issue is automatically linked and closed when this PR is merged. See https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

zeileis:
I hope I have correctly incorporated your second point.

And regarding the first point: When the number of rows and columns is different, then certain combinations of `facet` specifications do not work. For example, this looks as intended:

    tinyplot(mpg ~ wt, facet = vs ~ gear, data = mtcars)



But this is messed up because the layout is still 2 x 3 rather than 3 x 2 and hence it is not surprising that the annotation is not correct:

    tinyplot(mpg ~ wt, facet = gear ~ vs, data = mtcars)



If you have more variables in the `facets` that are then interacted, things can get even worse.

grantmcdermott:
Ah, I see now.

Thanks, let me take a deeper look. Hopefully, the fix won't be too difficult...

grantmcdermott:
Hi @zeileis. I've pushed a a few commits to your PR&mdash;hope that's okay&mdash;that seem to have plugged the problem areas. Quick examples...

``` r
#> ℹ Loading tinyplot
tpar(las = 1)
```

This was working beforehand and still does (2 x 3 arrangement):

``` r
tinyplot(mpg ~ wt, facet = vs ~ gear, data = mtcars)
```



This one wasn't working correctly beforehand, but now does (3 x 2 arrangement)):

``` r
tinyplot(mpg ~ wt, facet = gear ~ vs, data = mtcars)
```



And here's a more complicated example with multiple variables in the facet formula LHS , which used to trigger an error due to missing values in some facet windows, but now plots everything correctly. 

``` r
tinyplot(mpg ~ wt, facet = gear:vs ~ cyl, data = mtcars, pch = 16, axes = "l", grid = TRUE)
```



Compare against equivalent ggplot2 call:

``` r
library(ggplot2)
ggplot(mtcars, aes(wt, mpg)) +
  geom_point() +
  facet_grid(vs+gear~cyl, labeller = labeller(.multi_line = FALSE)) +
  theme_minimal()
```



<sup>Created on 2024-08-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Given that you're on a (well-deserved!) vacation, I imagine that it may be tricky for you to test on your side. But let me know what you think.

zeileis:
Wonderful, thanks a lot. I'll try to have a look tonight!

zeileis:
Very nice, thanks for these fixes! I've played around a bit and didn't find anything else. So IMO you can move forward and merge this.

Just one comment and one question:

- _Comment:_ Due to the new formula processing I think it's easier to use `facet = gear + vs ~ cyl` rather than `facet = gear:vs ~ cyl`. Both produce equivalent output but the former creates the `interaction()` internally rather than bothering the user with this. That's also what I would advertise in the examples.
- _Question:_ I noticed that in your last commit you used

      if (is.factor(facet)) {
        facets = levels(facet)
      } else {
        facets = sort(unique(facet))
      }

  rather than the more compact

      facets = if (is.factor(facet)) levels(facet) else sort(unique(facet))

  If the former is preferred, I can also adapt my code for this PR, e.g., https://github.com/grantmcdermott/tinyplot/blob/formula-terms/R/tinyplot.R#L1106 and https://github.com/grantmcdermott/tinyplot/blob/formula-terms/R/tinyformula.R#L23 but there are actually quite a few other constructs like this.

grantmcdermott:
Agree with both! The `if...{} else...{}` construct is mostly a function of habit from my side. I'll revert to your concise suggestion and then merge.

grantmcdermott:
Thank you @zeileis. This is great!

PR #214, opened 2024-09-09
Modularize: Minor PR

vincentarelbundock:
A minor PR with some more modularization/simplification:

* Removed duplicated code
* Removed unused arguments
* `type_*()` data preparation function are now stored in a dictionary with switch-style selection.


R/by_aesthetics.R
```diff
@@ -247,3 +247,47 @@ by_lwd = function(ngrps, type, lwd=NULL) {
...
+  by_continuous,
+  by_ordered,
+  col,
+  fill,
+  ngrps,
+  palette,
+  ribbon.alpha,
+  type) {
+
+  if (is.null(bg) && !is.null(fill)) bg = fill
```
grantmcdermott, line 266: Should this go higher up in the main tinyplot function? That way, we can also drop the extra fill argument for this `by_bg` function (as well as the input on l. 665 below).

grantmcdermott (changes requested):
Very nice, thanks. One minor comment and then happy to merge.

grantmcdermott (approved):

grantmcdermott:
Not so minor in the end... thanks again for the continued efficiency gains!

vincentarelbundock:
no it really was minor. I just noticed that your suggestion could help us simplify something else too.

This was a simple PR, but I'll take the 35 lines of code removed :)

PR #216, opened 2024-09-11
Support axis flipping

### LINKED ISSUE #184: flip axes
It would be cool if we could support a generic `tinyplot(..., flip = <logical>`) argument that flips the x and y axis orientation (similar to `coord_flip()` in ggpot2).

### LINKED ISSUE #223: Horizontal (flipped) boxplot axis titles need to be swapped too
Note the wrong axis titles below.

``` r
tinyplot::plt(Petal.Length ~ Species, iris, horizontal = TRUE)
```

<sup>Created on 2024-10-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

(First noticed following @zeileis's comment [here](https://github.com/grantmcdermott/tinyplot/pull/216#issuecomment-2395702628). )

Not sure how I missed it originally... but will push a fix as part of the same PR.

### LINKED ISSUE #224: Ribbon types without `ymin`/`ymax` should inherit from `y`
Otherwise you get nonsensical plots like:

``` r
tinyplot::plt(Petal.Length ~ Sepal.Length, data = iris, type = "ribbon")
```

<sup>Created on 2024-10-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Another option would be to error out explicitly with warning that `ymin` and `ymax` are required args for `type = "ribbon"`. But I think inheriting from `y` (so that the above plot looks the same as a line plot) is a better default, since we already draw `y` as a line if it is provided.

(H/T @zeileis comment [here](https://github.com/grantmcdermott/tinyplot/pull/216#issuecomment-2396182586). I'll fix as part of the same PR.)

grantmcdermott:
Closes #184.
Closes #223. (Bonus)
Closes #224. (Bonus)

This PR adds a `flip` argument that works with all plot types. @vincentarelbundock @zeileis I'd be very grateful if either if you could take a take look.

Some quick examples adapted from the new tests:

``` r
#> ℹ Loading tinyplot

op = tpar(las = 1, grid = TRUE)

tinyplot(
  ~ Petal.Length | Species, data = iris,
  type = "density", fill = 0.5,
  flip = TRUE,
  main = "Flipped density"
)
```

``` r

m = transform(mtcars, gear = factor(gear))
mod = lm(mpg ~ wt*gear, m)
coefs = data.frame(names(coef(mod)), coef(mod), confint(mod))
coefs = setNames(coefs, c("term", "estimate", "lwr", "upr"))

tpar(cex.axis = 0.75)

with(
  coefs,
  tinyplot(
    term, estimate, ymin = lwr, ymax = upr,
    type = "pointrange",
    pch = 19,
    flip = TRUE,
    axes = "l", xlab = NA,
    main = "Flipped coefplot"
  )
)
abline(v = 0, lty  = 4, col = "hotpink")
```

``` r

tpar(op)
```

<sup>Created on 2024-10-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

P.S. I also managed to simplify a few internal elements (e.g., removing unused arguments in some internal functions). But that's on the minor side.


grantmcdermott:
This PR adds a `flip` argument that works with all plot types. @vincentarelbundock @zeileis I'd be very grateful if either if you could take a take look.

Some quick examples adapted from the new tests:

``` r
#> ℹ Loading tinyplot

op = tpar(las = 1)

tinyplot(
  ~ Petal.Length | Species, data = iris,
  type = "density", fill = 0.5,
  grid = TRUE,
  flip = TRUE,
  main = "Flipped density"
)
```



``` r

m = transform(mtcars, gear = factor(gear))
mod = lm(mpg ~ wt*gear, m)
coefs = data.frame(names(coef(mod)), coef(mod), confint(mod))
coefs = setNames(coefs, c("term", "estimate", "lwr", "upr"))
tpar(cex.axis = 0.75)
with(
  coefs,
  tinyplot(
    term, estimate, ymin = lwr, ymax = upr,
    type = "pointrange",
    pch = 19,
    flip = TRUE,
    grid = TRUE, axes = "l", xlab = NA,
    main = "Flipped coefplot"
  )
)
abline(v = 0, lty  = 4, col = "hotpink")
```



``` r

tpar(op)
```

<sup>Created on 2024-10-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

P.S. I also managed to simplify a few internal elements (e.g., unused arguments in some internal functions). But that's on the minor side.

NEWS.md
```diff
@@ -6,11 +6,20 @@ where the formatting is also better._
 
 ## 0.2.1.99 (dev version)
 
+New features:
+
+- The new `flip` argument allows for easily flipping (swapping) the orientation
+of the x and y axes. This should work regardless of plot type, e.g.
+`plt(~Sepal.Length, data = iris, type = "density")`. (#216 @grantmcdermott)
```
vincentarelbundock, line 13: Where's the `flip` argument in this call?

R/draw_elements.R
```diff
@@ -22,36 +22,46 @@ draw_elements = function(
...
       if (isTRUE(empty_plot)) {
 
       } else if (type == "ribbon") {
-        polygon(
-          x = c(xx, rev(xx)),
-          y = c(yymin, rev(yymax)),
-          col = bg[i],
-          border = FALSE
-        )
+        if (isFALSE(flip)) {
```
vincentarelbundock, line 33: I recommend adding an assertion at the very top of the function `assert_flag(flip)`. This function is already in the package (but not used yet). It will ensure `TRUE`/`FALSE` and return an informative error. Otherwise, you probably want to change this condition because graphs will be flipped whenever someone passes a string or other non-conforming value.

vincentarelbundock:
Looks really great! I don't have time for a deep dive, but looked over quickly and left a couple comments. Very cool feature.

NEWS.md
grantmcdermott: D'oh. good catch, thanks.

R/draw_elements.R
grantmcdermott: Ah, nice. I'll add it to main `tinyplot.R` function since we already call `isTRUE(flip)`/`isFALSE(flip)` upstream there.

vincentarelbundock:
Out of curiosity, what is the motivation for `flip`, rather than just asking the user to swap x and y?

Wouldn't it be more explicit to allow this?

`plt(y = somevector, type = "hist")`

grantmcdermott:
How would you draw a sideways histogram or density plot with the existing codebase? ;-)

More generally, sure, swapping x and y directly suffices for some plot types (e.g., points). But for others it doesn't yield sensible b/c of the way that elements are drawn (e.g., polygon ordering).

vincentarelbundock:
Maybe we're just smart about it and apply an internal flip when people try a y histogram? That might be one our magical features.

polygon might be a real problem, but I haven't played enough with those to get a sense.

grantmcdermott:
Magic behavior like you describe would be nice, but likely a maintenance pain on our side... and potentially confusing to the user too, I think.

E.g. With this `flip` approach, you only need to change (add) one argument to get the flipped coefplot above. If we were doing magic inference, then the user would need to change four arguments and we'd also need to sanity check on the backend that no duplicate xmin/ymin or xmax/ymax arguments had been provided.

It's quite similar to `coord_flip` in ggplot2, with the exception that, once the initial plot is drawn, new elements can be added using the actual visual orientation (e.g., `abline(v = 0)`) instead of pretending that we're still in the original orientation (e.g., `abline(h = 0)`). The latter is something I always find a bit annoying about coordinate flipping in ggplot2, especially if I am drawing plots via a wrapper function.

vincentarelbundock:
Cool cool. I'm convinced

zeileis:
Nice, feature, thanks! I also agree that adding this one argument will make life easier for plots where x and y cannot simply be flipped. For example, `numeric ~ factor` gives a different base `plot()` than `factor ~ numeric`. Yet flipping the axes for these may be useful in some settings.

I just started playing around and noticed that the default grid works well in

```
tinyplot(Petal.Length ~ Species, data = iris)
```

but is off in

```
tinyplot(Petal.Length ~ Species, data = iris, flip = TRUE)
```

I'll play around a bit more after some sleep...

zeileis:
I didn't look at the details of the code but just tried a range of different plot with `flip = FALSE` vs. `flip = TRUE`. This all looks great and very useful. In addition to the grid problem reported above I could only find the following issue with a non-sensical example where the error handling is apparently different:

```
tinyplot(Petal.Length ~ Sepal.Length, data = iris, type = "ribbon")
tinyplot(Petal.Length ~ Sepal.Length, data = iris, type = "ribbon", flip = TRUE)
```

grantmcdermott:
Thanks @zeileis!

Both of those actually alerted me to ancillary issues, for which I've rolled fixes into this PR.

PR #219, opened 2024-09-17
Coerce x/y/by from character to factor

### LINKED ISSUE #217: Should we support `plt(numeric ~ character)`?
And vice versa.

Currently, this fails (as it would for regular `plot` too).

```r
dat = data.frame(x = letters[1:10], y = 1:10)
tinyplot::plt(y ~ x, dat)
```

Maybe we should coerce characters to factors internally?

zeileis:
Fixes #217

**Summary:** In `tinyplot.default` the `x`, `y`, and `by` variables are turned into factors if they are character variables. This is done early on in the function but after all `deparse1(substitute(...))` calls because the coercion would evaluate the variables obviously.

**Caveat:** In the modular future we might have tinyplot types which want to draw character variables in some way. As long as the type "knows" that, it's not yet a killer argument. The plot type could coerce back with `as.character()` and end up with the original variable. However, if that plot type would do somewhat different things for character variables and factor variables, then the design decision from this PR would prevent that.

### DIFF
--- a/R/tinyplot.R
+++ b/R/tinyplot.R
@@ -587,9 +587,14 @@ tinyplot.default = function(
   }
   by_dep = deparse1(substitute(by))
   
+  ## coerce character variables to factors
+  if (!is.null(x) && is.character(x)) x <- factor(x)
+  if (!is.null(y) && is.character(y)) y <- factor(y)
+  if (!is.null(by) && is.character(by)) by <- factor(by)
+
   # flag if x==by (currently only used if type = "boxplot")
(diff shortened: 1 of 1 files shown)


grantmcdermott:
Thanks @zeileis!

(Also for raising the caveat. Good to keep in mind and might necessitate a `was_character` flag or similar down the line.)

R/tinyplot.R
zeileis, line 1: Sorry, I'll try to remember the next time...

PR #221, opened 2024-09-21
lim_args was updating xlim and ylim even if specified

mclements:
Nice package -- great to see base R graphics getting some attention.

This is a very minor bug fix. The following code shows that `xlim` and `ylim` were being ignored when ymin/etc are specified:

```
library(tinyplot)
x=y=0:1
conf.low=y-0.5
conf.high=y+0.5
par(mfrow=1:2)
plt(y~x,ymin=conf.low,ymax=conf.high,type="ribbon")
plt(y~x,ymin=conf.low,ymax=conf.high,type="ribbon",xlim=c(0,0.5),ylim=c(0,0.5))
```

The code changes `lim_args()` to check whether `xlim` or `ylim` are `NULL`.


vincentarelbundock:
Thanks for taking a look at this!

Looks like it breaks some plots. In particular, it cuts the confidence intervals.

To run the test suite and check this yourself, you can comment out the "Linux" line in `inst/tinysnapshot/helpers.R`. Then,

```{r}
pkgload::load_all()
tinytest::run_test_dir()
```

ribbon


pointrange_triangle

mclements:
Interesting - I **want** to cut the confidence intervals, particularly when the intervals are very wide. Semantically, I suggest that `xlim` and `ylim` should be followed when specified (as per `base::plot()`). Do you disagree?

mclements:
I have further simplified the calculations for `xlim` and `ylim`. 

I am getting errors for `test-legend-gradient.R` (6/7) and `test-rect.R` (1/2) due to different versions of libpng -- which may be due to how the test cases were generated. 

Can you re-run the tests, please?

vincentarelbundock:
Thanks a lot for this @mclements ! It's a great and useful bug fix.

I updated a few snapshots (no visual difference at all) and everything passes. Merging now.

Cheers,

Issue #236, opened 2024-10-29
Adding `abline`s to faceted plots


grantmcdermott:
Adding an `abline` to a (simple) `tinyplot` works much the same as base `plot()`.

``` r
library(tinyplot)
library(marginaleffects)

fit = lm(Sepal.Length ~ (Sepal.Width + Petal.Length + Petal.Width) * Species, iris)

mfx = avg_slopes(fit, variables = c("Sepal.Width", "Petal.Length", "Petal.Width"))
with(
  mfx,
  plt(
    x = term, y = estimate, ymin = conf.low, ymax = conf.high,
    type = "errorbar", pch = 19
  )
)
abline(h = 0, lty = 2, col = "hotpink")
```

![](https://i.imgur.com/ZEW4ZY4.png)

So far so good. But unfortunately this approach doesn't work for faceted plots:

``` r
mfx2 = avg_slopes(fit, variables = c("Sepal.Width", "Petal.Length", "Petal.Width"), by = "Species")
with(
  mfx2,
  plt(
    x = term, y = estimate, ymin = conf.low, ymax = conf.high,
    type = "errorbar", pch = 19,
    facet = Species, facet.args = list(bg = "gray90"), frame = FALSE
  )
)
abline(h = 0, lty = 2, col = "hotpink")
```

![](https://i.imgur.com/DONpZCo.png)

I don't see a good way to support the standard way of adding of `abline`s to faceted plots post hoc. (Too many complications.) But we could potentially support an  `abline` _argument_ that would draw the required line during each facet, and similarly for single facet plots too. Something like:

```r
tinyplot(..., abline = list(h = 0))
```

Thoughts

vincentarelbundock:
Is this too weird?

`plt(type=type_abline(), add=TRUE)`

Maybe...

zeileis:
I wouldn't add another specialized argument like `abline` to `tinyplot()`.

Either we should go via the route that Vincent suggested, i.e., a separate `tinyplot(..., add = TRUE)`.

And/or, alternatively, we could add a `panel = ...` argument that can draw stuff within each facet of a plot. The crucial question is then which information this panel function has access to and what arguments it can take. Also, this might have to control what is drawn before and what after the `type_draw`.

grantmcdermott:
> Is this too weird?
> plt(type=type_abline(), add=TRUE)

[...]

PR #246, opened 2024-11-08
`plt_add()`

vincentarelbundock:
I’ve been trying to think about ways to make it easier to layer plots. We have `add=TRUE`, but it is not very convenient, because we need to repeat all the same arguments in every call.

I can’t decide if this is a good or a terrible idea, but here’s a proof of concept for you to comment on:

1.  Every time we do `plt()`, we save the full call somewhere (global option or other).
2.  New function `plt_add()`
    - Takes the last call,
    - Modifies the explicitly-given argument
    - Sets `add=TRUE`
    - Evaluates the modifed call

This proof of concept PR allows things like:

``` r
library(tinyplot)

plt(Sepal.Width ~ Sepal.Length | Species,
    data = iris, 
    type = type_lm())

plt_add(type = "p")
```

Note: Facets don’t seem to work in this PR.


grantmcdermott:
This was exactly what I was thinking of in the digression here: https://github.com/grantmcdermott/tinyplot/issues/236#issuecomment-2444939778

Great minds etc. ;-)

I really like the idea and think that this is a feature that would get a lot of use. I'll be able to take a look at it tomorrow (or whenever you move it out of draft).

vincentarelbundock:
@grantmcdermott I added docs and a test. You might want to look at this now.

R/tinyplot.R
```diff
@@ -536,6 +536,10 @@ tinyplot.default = function(
     ...
     ) {
 
+  # save for plt_add()
+  options(tinyplot_last_call = match.call(tinyplot,
```
grantmcdermott, line 540: To be consistent with the rest of the package infrastructure, let's rather save this to the existing `.tinyplot_env`. This is what we use in `R/get_saved_par.R`, for an example:

https://github.com/grantmcdermott/tinyplot/blob/02d78063aa25aef004a49357b9632c71957037b3/R/get_saved_par.R#L104-L116

We'll need to assign NULL values to this call object at startup in `R/zzz.R` too, e.g.
https://github.com/grantmcdermott/tinyplot/blob/02d78063aa25aef004a49357b9632c71957037b3/R/zzz.R#L46

vignettes/introduction.qmd
```diff
@@ -438,6 +438,27 @@ tinyplot(
 The `facet.args` customizations can also be set globally via the `tpar()`
 function, which provides a nice segue to our penultimate section.
 
+## Layers
+
+In many contexts, it is convenient to build plots step-by-step, adding layers with different elements on top of a base. 
+
+The `tinyplot()` function includes an `add=TRUE` argument that adds element to an existing plot, instead of drawing a new window. This argument is useful, but a bit verbose, as it requires users to make very similar successive calls, with many shared arguments.
+
+For convenience, the package also includes a `tinyplot_add()` function which captures the last `tinyplot()` call, keeps all the same arguments, and modifies just the arguments that the user explicitly wants to change.
```
grantmcdermott, line 447: Can we mention the `plt_add()` shorthand alias here too?

vignettes/introduction.qmd
```diff
@@ -438,6 +438,27 @@ tinyplot(
...
+
+```{r}
+library(tinyplot)
+
+tinyplot(Sepal.Width ~ Sepal.Length | Species,
+         facet = ~Species,
+         data = iris, 
+         type = type_lm())
+
+tinyplot_add(type = "p")
```
grantmcdermott, line 459: Just checking: Do the y-axis limits spacing look good in this example? I recall having to draw the points first and then adding the lm fit afterwards in the manual `add = TRUE` example. (That, or manually setting `ylim` in the first call.)

R/tinyplot_add.R
```diff
@@ -0,0 +1,44 @@
...
+#'          data = iris, 
+#'          type = type_lm())
+#' 
+#' tinyplot_add(type = "p")
+#'
+#' @returns No return value, called for side effect of producing a plot.
+#'
+#' @export
+tinyplot_add <- function(...) {
+  cal = getOption("tinyplot_last_call")
```
grantmcdermott, line 28: See note about `.tinyplot_env` above.

vignettes/introduction.qmd
grantmcdermott: Just checked; I called lm first but set `ylim`. On that note, should this new `## Layers` section replace the code that I have here?

https://github.com/grantmcdermott/tinyplot/blob/02d78063aa25aef004a49357b9632c71957037b3/vignettes/introduction.qmd#L401-L418

grantmcdermott (changes requested):
Looks great. I requested some changes that should be relatively minor to action hopefully.

R/tinyplot.R
vincentarelbundock: Unfortunately, I was unable to get this to work. I kept getting an error in the `tinyplot.density` tests. The message is obscure, having to do with a `do.call` and a `substitute`. I just can figure this out.

My guess is that this will disappear by itself when we refactor the density code.

In the meantime, I left TODOs in both files with the appropriate save code that we can eventually.

Feel free to take a shot at it if you want, but I have to give up for now.

grantmcdermott:
This is great. Thanks @vincentarelbundock!

PR #249, opened 2024-11-12
Types: `abline`, `hline`, `vline`

vincentarelbundock:


vincentarelbundock:
I feel like the whole refactor is paying off in a really big way now :)

altdoc/quarto_website.yml
```diff
@@ -36,12 +35,15 @@ website:
             file: man/draw_legend.qmd
         - section: Plot types
           contents:
-          #
           - section: Shapes
             contents:
-            - text: type_area, type_ribbon
```
grantmcdermott, line 42: I think this change should be revert since area plots are a special case of ribbon plots. The underlying docs point to the same help file.

R/type_abline.R
```diff
@@ -0,0 +1,72 @@
+#' Add straight lines to a plot
+#'
+#' @inheritParams graphics::abline
+#' @inheritParams tinyplot
+#' @examples
+#' mod = lm(mpg ~ hp, data = mtcars)
+#' y = mtcars$mpg
+#' yhat = predict(mod)
+#' tinyplot(y, yhat, xlim = c(0, 40), ylim = c(0, 40))
+#' tinyplot_add(type = type_abline(a = 0, b = 1))
```
grantmcdermott, line 10: This is great. Just so I know: What happens if you can `tinyplot(..., type = type_abline()` directly (i.e., without adding to an existing plot)? Do we need to add a `stop()` clause or warning for users that try to do this?

grantmcdermott (changes requested):
Nice! I didn't have time to test this locally, but just left a few minor comments.

R/type_abline.R
vincentarelbundock: good catch. I added a `stop()` logic and some tests

altdoc/quarto_website.yml
vincentarelbundock: I can revert if you like. I originally changed it because it hurts readability in the side bar. And since we're going to have *a lot* of types...

Screenshot 2024-11-13 at 6 53 22 AM

grantmcdermott (approved):

grantmcdermott:
Brilliant, thanks!

I've got some pressing deadlines at work. But will try to take a look at your other PRs later today.

PR #262, opened 2024-11-20
intro plt() alias discussion

vincentarelbundock:
The `plt()` tooltip is great and highly visible.

I moved it to just after we call `tinyplot()` for the first time, so users see it immediately.

I removed the section on `plt()`. IMHO, the tooltip is already very visible, and spending a full section on an alias at the top of the intro vignette is overkill. The start of the introduction in prime real estate in a tutorial, and we should invest it on "real" features.

### DIFF
--- a/vignettes/introduction.qmd
+++ b/vignettes/introduction.qmd
@@ -63,21 +63,8 @@ earlier.
 tinyplot(Temp ~ Day, data = aq)  # formula method
 ```
 
-### Aside: `plt` shorthand
-
-If you'd prefer to save on a few keystrokes, you can use the shorthand `plt`
-alias instead of typing out `tinyplot`.
-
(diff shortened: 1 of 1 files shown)


grantmcdermott:
You make a good point. But, in that case, should we just go all the way and move the tooltip to the very top of the document? (I.e. Just after the main heading.)

vincentarelbundock:
Well I think the main function should be presented before its alias. Otherwise we can't even answer: "alias to what?"

But anywhere after the first evaluated `tinyplot()` call  is fine, I think.

grantmcdermott:
Circling back to this with a few tweaks. If you're happy with my suggestion, please feel free to merge. (Or let me know and I will.)

vincentarelbundock:
LGTM

Issue #265, opened 2024-11-21
type_abline deprecate aesthetics


vincentarelbundock:
Currently type_abline() and friends all accept their own aesthetics like col and lty.

These arguments should not be in the type function. They should be governed by the top level tinyplot()

PR #266, opened 2024-11-22
Issue #265 `col`, `lty` and friends should not be `type_abline()` arguments

vincentarelbundock:


vincentarelbundock:
Very uncontroversial. Just *removing* code, not changing much.

Tests pass. Merging now. (Let me know if I should refrain from things like this in the future.)

PR #270, opened 2024-11-27
ylevels reordering in type_ridge()

### LINKED ISSUE #271: Density scaling and bandwidth in type_ridge()
In `get_density()` within `type_ridge()` the density estimates are scaled to the same maximum which I believe is incorrect. Also the bandwidths are by default computed separately whereas `ggridges` employs a joint bandwidth estimate which typically seems to be better. But maybe I'm overlooking something here, Vincent @vincentarelbundock ?

For illustration let's consider a sample with two groups with very different variances:

```
set.seed(0)
d <- data.frame(
  y = rep(1, 100),
  x = c(rnorm(50), rnorm(50, mean = 3, sd = 0.2)),
  z = factor(rep(1:2, each = 50))
)
```

Using `ggridges` a joint bandwith of 0.218 is selected and the resulting densities have very different maxima - as you would expect if both should have the area 1.

```
ggplot(d, aes(x = x, y = z)) + geom_density_ridges() + theme_minimal()
ggplot(d, aes(x = x, y = y, group = z)) + geom_density_ridges() + theme_minimal()
```

tinyplot-ridge1

In contrast, because `get_density()` scales the density to have the same maximum, the second group has a much smaller area than the first in `tinyplot(..., type = "ridge")`.

```
tinyplot(z ~ x, data = d, type = "ridge", grid = TRUE)
tinyplot(y ~ x | z, data = d, type = "ridge", grid = TRUE)
```

tinyplot-ridge2

However, if I drop the scaling in line https://github.com/grantmcdermott/tinyplot/blob/main/R/type_ridge.R#L99 and select the same bandwidth of 0.218, then I get virtually the same result as in `ggridges`.

```
tinyplot(z ~ x, data = d, type = type_ridge(bw = 0.218), grid = TRUE)
tinyplot(y ~ x | z, data = d, type = type_ridge(bw = 0.218), grid = TRUE)
```

tinyplot-ridge3

Note that `type = "density"` agrees here and chooses the same joint bandwidth without applying any scaling of the maximum:

```
tinyplot(~ x | z, data = d, type = "density", grid = TRUE)
```

tinyplot-ridge4

zeileis:
Fixes #271

This is a mini PR that re-orders the categories of the y-variable in `type_ridge()` as in `type_spineplot()`. This is done in the `type_data` function at the beginning (after determining whether `y` is equal to `by`).

_Edit:_ In addition to the `ylevels` I will address the problems with density scaling and bandwidth here.


grantmcdermott:
Thanks for doing this (and fixing the density scaling/bandwidth issue too). I haven't looked into why the tests are failing, but one meta question I have is the following:

Shouldn't re-ordering of the y- (or x-) axis be handled generically by the main `ylim`/`xlim` arguments? For example, I've wondered whether we shouldn't support a way to easily reverse the direction of an axis _a la_ [this example](https://r-graph-gallery.com/77-turn-y-axis-upside-down.html), say, `tinyplot(..., ylim = rev)`.

zeileis:
The tests are expected to fail because I already changed the scaling of the densities and thus all ridge type plots now look different than before.

For changing the bandwidth I'm not sure what the best approach would be. Any guidance would be helpful, see #271.

The arguments `ylevels` and `ylim` are separate aspects. The former is about re-ordering the categories of the response. The latter modifies the range actually drawn. In a spine plot you might want to have a subset of [0, 1] and in a ridge line you might also want to zoom in or out.

grantmcdermott:
Got it, thanks for the explanation.

> For changing the bandwidth I'm not sure what the best approach would be. Any guidance would be helpful, see #271.

I'm neck deep in the themes PR rn, but will take a look as soon as I can after.

zeileis:
Sure, no worries. My time will be very limited in the next two weeks due to other projects anyway. I'm optimistic, though, that we'll finish it as a nice Christmas present to ourselves 😅

grantmcdermott:
Sorry for the long delay in reviewing this @zeileis. Am I correct in thinking that this where we stand?

- The PR fixes both the y-levels and density scaling issue for type_gradient?
  - However, it doesn't implement the `weighted.mean` suggestion for selecting a common bandwidth both for `type_gradient` and across regular grouped `type_density()` plots.
- We still need to delete the old gradient snapshots and replace them with the fixed versions so that the tests pass.

zeileis:
Status:

- `ylevels` is done. 
- Scaling is definitely better than before and necessitates updating all corresponding snapshots. We can leave it at this (for now) or we can think about adding something like `panel_scaling` in `geom_density_ridges()`. This controls whether the scaling is done per panel (default) or globally.
- The bandwidth I did not touch, yet, because I wasn't sure what was the best way to do it. So this is still open and `weighted.mean()` seems to be our current favorite solution.

grantmcdermott:
Great, thanks for the update. I cloned your branch locally and tried a few of the snapshots. I agree that the new scaling makes more sense as a default.

> We can leave it at this (for now) or we can think about adding something like `panel_scaling` in `geom_density_ridges()`. This controls whether the scaling is done per panel (default) or globally.

I think that this is a good idea.  Do we have a better argument name than `panel_scaling`, though? I don't have hugely strong priors, but it doesn't feel like the most intuitive name to me.

grantmcdermott:
P.S. I'm going to push the updated snapshots (alongside merging in main), just so that these are up to date with what the PR is actually proposing.

grantmcdermott:
Okay, I've pushed my proposed changes as discussed in https://github.com/grantmcdermott/tinyplot/issues/271#issuecomment-2611648994. Apologies: the git diff makes it look like quite a lot has been done to the code... But in truth much of this is purely cosmetic since I wanted to align the `type_density()` and `type_ridge()` internal code bases for easier maintenance going forward. The material changes aren't that significant.

Alongside Achim's changes from earlier, the key points of my additions:

- Like `type_density()`, `type_ridge()` now accepts a `joint.bw` argument for specifying a joint bandwidth calculation. The key difference is that the current default case here is `type_ridge(joint_bw = "owm")` (i.e., observation-weighted mean), whereas we agreed to a default of `type_density(joint.bw = "none")` (i.e., no joint bandwidth) in #284.
   - **Key question for @zeileis and @vincentarelbundock:** Is everyone agree with this discrepancy or do we  we want to enforce the same default between `type_density()` and `type_ridge()`? My own feeling: I'm fine with the discrepancy, but could live with a shared default of `"owm"` if we really want to enforce consistency. (I really don't like the `"none"` look for ridge plots, so wouldn't want that as a shared default.) FWIW, `ggridges::geom_density_ridges()` deviates from `geom_density()`, so we'd at least be in reasonable company by not enforcing consistency across these types.
- I've also fixed some documentation and rendering issues that I caught during this review (e.g., advice for combining gradient coloring with alpha transparency).

grantmcdermott:
@zeileis @vincentarelbundock Sorry to keep pestering, but I'm really hoping to resolve this density vs ridge `joint.bw` issue so we can submit v0.3.0 to CRAN.  If you won't have time to review or think about it in the next one or two days, absolutely no worries. But please just let me know and then I'll make an executive decision.

Let me also explicitly state the options again, since it might have gotten buried in the back and forth..

### Option 1: Different joint.bw calculations for density and ridge types

As things stand at the moment, this PR will introduce a discrepancy between `type_density()` and `type_ridge()` i.t.o. their default bandwidth behaviour for grouped/multi density curves. Specifically, the default for `type_density()` is "none" (i.e., no joint bandwidth), whereas for type_ridge()` it is "owm" (observation-weighted mean). Quick example:

```r
plt(~ Petal.Width | Species, data = iris, type = type_density())   # default joint.bw = "none"
plt(Species ~ Petal.Width, data = iris, type = type_ridge())       # default joint.bw = "owm"
```

density
ridge

Personally, I'm _okay_ with this discrepancy. And it  would at least be consistent with what `geom_density()` and `geom_density_ridges()` do since they also diverge in their bandwidth behaviour. But we obviously don't have to follow their choices, which leads to me to option 2...

### Option 2: Use same joint.bw calculation for both density and ridge types

If we'd like the ensure consistency in defaults, then we could use "owm" for both. For the current example, that would effectively look like:

```r
plt(~ Petal.Width | Species, data = iris, type = type_density(joint.bw = "owm"))
plt(Species ~ Petal.Width, data = iris, type = type_ridge())
```
density-owm
ridge

Currently, I'm slightly leaning towards this option. I like the internal consistency and the the level of smoothing is more aesthetically pleasing (although I recognize that it can also be misleading).

WDYT?

vincentarelbundock:
I have the weakest of preferences on this issue, so feel free to make an executive decision.

FWIW, I am also slightly leaning toward the latter because I think users like consistency.

The issues discussed in the linked comments are of course valid, but since we now have a good way to document the options, users can easily switch to another strategy if they want. So that seems less pressing that it may have once been.

But again, weak preference.

I'm just really looking forward to using `tinyplot` "full time", and I think version 0.3.0 will finally get me there.

grantmcdermott:
Okay, executive decision: I'm going to go for **Option 2** :-)

Achim, if you're reading this after the fact, sorry not giving you time to respond. But I've got a (rare) free hour or two to make the necessary changes and want to strike while the iron is hot... and the rest of the family is out of the house!

I'll merge this PR as-is and then address consistency for `type_density()` in a separate PR to maintain a proper paper trail.

zeileis:
Sigh, was working on it right now...

grantmcdermott:
Ugh, sorry @zeileis!

Why don't you just post them here regardless and I'll think on the best way to merge them.

zeileis:
OK, I'll probably post a few comments here and put other things in a new PR from the now main branch. If you don't like them, you can still discard them. Need a bit more time to try out more variations. This is all surprisingly complicated and depending on how much overlap there is between the different groups.

zeileis:
Your examples have convinced me that `joint.bw = "owm"` is a pragmatic choice for both `type_density()` and `type_ridge()`. In #291 I propose some improvements in the documentation and the renaming of the opaque `"owm"` to `"mean"`.

One additional quibble I still have is the `global.max` option. I don't understand when `global.max = FALSE` makes sense as the areas of the different densities will not be the same anymore. Is this done in other ridge plot implementations?

What might be useful, however, is to compute separate maxima for each facet. If I recall correctly, this is what `ggridges` does. In that case, areas between facets may not be comparable - but they are still within each facet.

grantmcdermott:
> One additional quibble I still have is the `global.max` option. I don't understand when `global.max = FALSE` makes sense as the areas of the different densities will not be the same anymore. Is this done in other ridge plot implementations?

Hmmm. I think it's potentially useful as a representation, even if the areas aren't strictly comparable. I've used it in summary figures before, for example, when quickly comparing groups across y categories.

```r
tinyplot(
  cyl ~ mpg | factor(am), data = mtcars,
  type = type_ridge(scale = 1, global = FALSE),
  width = 8, height = 5, file = "tmp.png"
)
```
tmp

_Caveat emptor_, I guess?

> What might be useful, however, is to compute separate maxima for each facet. If I recall correctly, this is what `ggridges` does. In that case, areas between facets may not be comparable - but they are still within each facet.

Yes, ggridges does do this (by default IIRC). We could probably add some logic that allows for this, although I wouldn't prioritise it right now.

zeileis:
Ah, that makes sense, thanks! If you contrast multiple densities on the same row you can standardize them like this. Then the main focus is on the comparison within the same `by` row and that works. However, if there are no `by` groups, then naturally the contrast is between rows and that does not work.

For illustration consider this simple artificial example where means and/or standard deviations differ between groups:

```
set.seed(0)
d <- data.frame(
  y = c(
    rnorm(100, mean = 0, sd = 0.2),
    rnorm(100, mean = 1, sd = 1),
    rnorm(100, mean = 2, sd = 0.35),
    rnorm(100, mean = 2, sd = 0.7)
  ),
  x = factor(rep(c("a", "b", "a", "b"), each = 100)),
  z = factor(rep(c(1, 1, 2, 2), each = 100))
)
```

Comparing the `by` groups within the same row then works well and it is not distracting that the area scaling between the rows differs:

```
tinyplot(z ~ y | x, data = d, type = "ridge", global.max = FALSE, scale = 1)
```

max1

However, if you have now `by` groups then this type of scaling is really ugly and misleading:

```
tinyplot(interaction(x, z) ~ y, data = d, type = "ridge", global.max = FALSE, scale = 2)
```

max2

So maybe it would be better to use a character argument rather than a logical `TRUE`/`FALSE` argument. It could take values:

- `"all"` (default): Maximum is taken over all groups, i.e., preserving direct comparability of all ridge lines.
- `"facet"` (the strategy in `ggridges`): Comparability within the same facet.
- `"by"` (as discussed above): Comparability within the same row.

And if no `facet` or `by` is specified respectively, then a warning could be issues, resorting to the default.

grantmcdermott:
> So maybe it would be better to use a character argument rather than a logical `TRUE`/`FALSE` argument. It could take values:
> 
>     * `"all"` (default): Maximum is taken over all groups, i.e., preserving direct comparability of all ridge lines.
> 
>     * `"facet"` (the strategy in `ggridges`): Comparability within the same facet.
> 
>     * `"by"` (as discussed above): Comparability within the same row.

Makes sense to me. I don't know if I'll time to implement rn, so please feel free to put in a PR if you have time.

zeileis:
OK, done in #292

PR #282, opened 2024-12-26
minor legend simplification

vincentarelbundock:
This is totally optional. Feel free to close if you don't think the code is clearer.

Mostly replaced `is.null()` checks with a `fill_null()` helper, and inserted some `isTRUE` to simplify checks.


grantmcdermott:
Cool idea. I wonder whether we should make this a general function and move to `utils.R`?

At least, I've been thinking along those lines since R.4.4.0 introduced the `L %||% R` [null coalescing operator](https://search.r-project.org/R/refmans/base/html/Control.html):

> The null coalescing operator %||% is a simple 1-line function: x %||% y is an idiomatic way to call
> 
>     if (is.null(x)) y else x
>                              # or equivalently, of course,
>     if(!is.null(x)) x else y 

Maybe we could do something as simple as:

```r
if (getRversion() <= "4.4.0" ) {
   `%||%` = function (x, y) if (is.null(x)) y else x
}
```
?

vincentarelbundock:
yep, seems to work. I just pushed this.

grantmcdermott:
Great, thanks. Let's wait for #284 and then merge this

grantmcdermott:
(Just a HU, I'm going to merge after our 0.3.0. CRAN submission, so we can start bundling a bunch of internal changes and documentation improvements in v0.3.1.)

PR #284, opened 2024-12-27
Add type_density() function

### LINKED ISSUE #194: Custom legend not working for tinyplot.formula(..., type = "density")
The legend here should be positioned _inside_ the plot frame...

``` r
library(tinyplot)

tinyplot(
  ~ Petal.Length | Species, iris,
  type = "density", fill = "by",
  legend = "topright"
)
```

... in contrast, the atomic method correctly adjusts and positions the legend.

``` r

with(
  iris,
  tinyplot(
    Petal.Length,
    by = Species,
    type = "density", fill = "by",
    legend = "topright"
  )
)
```

<sup>Created on 2024-08-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### LINKED ISSUE #243: `type_density()` using the new `type_data`/`type_draw` system
Density plots are coded differently than all other types.

I won't have time to convert them now, but opening this issue because it would eventually be a good idea to bring them in conformity.

grantmcdermott:
Closes #194
Closes #243

We might need some back and forth to get this over the line. I'm about to head out for a cabin break without my laptop, although I will have intermittent cell signal. Hopefully I can respond to general review questions.

Some high-level notes:

1. I've overhauled (simplified) the joint bandwidth logic, following on from our discussion in https://github.com/grantmcdermott/tinyplot/issues/243#issuecomment-2558188788. The _tl;dr_ version is that I've gone for option _**(B)** re-using the same bandwidth that we would obtain from the full dataset._ I'm not wedded to this approach, but it was the simplest and I wanted to get the ball rolling for discussion/review on this PR. It does end up breaking a lot of tests, though. (Sorry if that adds to the review noise.)

2. I decided to keep the `tinyplot.density` method, but only for simple cases, i.e., no groups or facets. Passing either groups or facets through this method now triggers an error with a prompt to rather use the dedicated `type = type_density()` / `type = "density"` approach.

3. There are a couple of TODO / FIXME notes still in the code that I _think_ we can resolve once this `type_density()` refactoring goes through. One is for the `tinyplot_add()` call [here](https://github.com/grantmcdermott/tinyplot/blob/5c6000f68c3a62650d0ee1005c73f5f33e7cd74a/R/tinyplot.R#L579-L590) (@vincentarelbundock I think this is you; please confirm if you can).


R/type_area.R
```diff
@@ -1,25 +1,27 @@
 #' @rdname type_ribbon
 #' @export
-type_area = function() {
+type_area = function(alpha = NULL) {
```
vincentarelbundock, line 3: Is it more transparent (and possible) to set the default to `alpha = tpar("ribbon.alpha")` in the function signature?

R/type_density.R
```diff
@@ -1,237 +1,132 @@
...
+#' @param kernel a character string giving the smoothing kernel to be used. This
+#'   must partially match one of `"gaussian"`, `"rectangular"`, `"triangular"`,
+#'   `"epanechnikov"`, `"biweight"`, `"cosine"` or `"optcosine"`, with default
+#'   `"gaussian"`, and may be abbreviated to a unique prefix (single letter).
+#'
+#'   `"cosine"` is smoother than `"optcosine"`, which is the usual 'cosine'
+#'   kernel in the literature and almost MSE-efficient. However, `"cosine"` is
+#'   the version used by S.
+#' @inheritParams type_ribbon
+#' @param indiv.bw Should the smoothing bandwidth be computed individually
```
vincentarelbundock, line 15: Is this a standard arg name? If not, I might reverse the order in case there are other `bw`-related arguments in the future. That would allow us to clump them together in alphabetical-looking order. Also, if I'm looking for bandwidth options, I'll probably look for "b" rather than "i"

R/type_density.R
```diff
@@ -1,237 +1,132 @@
...
+        datapoints$ymax = datapoints$y
+        datapoints$ymin = rep.int(0, nrow(datapoints))
+        
+        # flags for legend and fill
+        dtype = if (!is.null(bg)) "ribbon" else "l"
+        dwas_area_type = !is.null(bg)
+        
+        out = list(
+            ylab = ylab,
+            xlim = range(datapoints$x),
```
vincentarelbundock, line 119: Do we need to overwrite `xlim`? What if the user wants to show only a portion of the density? 

I understand we nay need to override `ylim` because `density()` outputs on a different scale, but curious about `x`.

vincentarelbundock: And what if users explicitly set `xlim`. Will this break things?

vincentarelbundock:
I left a few minor comments but don't have much to add. This looks really nice. Feels simpler and cleaner.

I don't have a view on default bandwidth selection (global vs. individual), but having the argument to control this is nice.

grantmcdermott:
Hi @zeileis just a gentle push about the joint bandwith issue above, since I see you're back from vacation ;-) Again, I've laid out four options [here](https://github.com/grantmcdermott/tinyplot/issues/243#issuecomment-2558188788) and for this PR I've gone with option (B). But user's can also opt into option (A) if they've like.

R/type_density.R
grantmcdermott: It's not a standard argument, but I'd say it's consistent with the naming convention of `density` arguments (e.g., the `warnWbw` argument). I'd also want to avoid potential S3 dispatch issues/confusion with a leading `bw.<some_arg>`...

R/type_density.R
grantmcdermott: Good point. I'll fix when I get a sec.

R/type_area.R
grantmcdermott: I decide to keep this as `NULL` for the moment to be consistent with `type_ribbon()` (and thus `type_area()` too). We can revisit all of these cases later if we want,

R/type_density.R
grantmcdermott: https://github.com/grantmcdermott/tinyplot/pull/284/commits/fe92a21c492a9623e89c59614a4bfdf88fb1d027

R/type_density.R
grantmcdermott: https://github.com/grantmcdermott/tinyplot/pull/284/commits/4bcb14491448dd0225cf1d2c5acd27586bc83d05

grantmcdermott: Perhaps not "resolved" to your complete satisfaction, but I've switched this out for a `joint.bw` argument now since I've gone with individual bandwidths by default (see my comment in the main thread).

grantmcdermott:
Thanks both. I've incorporated your suggestions and will merge once the checks pass.

One quick example: Based on our discussion in #243, the default for grouped densities is now to use individual bandwidths (also consistent with ggplot2). But users can override and select a joint bandwidth using either the full dataset or one computed as the observation-weighted mean.

``` r
tinyplot(~Sepal.Length | Species, data = iris, type = "density") # "none" (default)
tinyplot_add(type = type_density(joint.bw = "full"), lty = 2)    # full dataset
tinyplot_add(type = type_density(joint.bw = "owm"), lty = 3)     # obs-weighted mean
legend("topright", c("None", "Full", "OWM"), lty = 1:3, title = "Joint BW")
```

PR #292, opened 2025-01-28
type_ridge: joint.max=all/facet/by instead of global.max=TRUE/FALSE

zeileis:
Grant @grantmcdermott, this is a follow-up to our discussion in #270 

I replaced the old `global.max = TRUE` vs. `FALSE` argument with a new `joint.max = "all"` vs. `"facet"` or `"by"`. The documentation reads as follows.

`joint.max`: character indicating how to scale the maximum of the densities: The default `"all"` indicates that all densities are scaled jointly relative to the same maximum so that the areas of all densities are comparable. Alternatively, `"facet"` indicates that the maximum is computed within each facet so that the areas of the densities are comparable within each facet but not necessarily across facets. Finally, `"by"` indicates that each row (in each facet) is scaled separately, so that the areas of the densities for `by` groups in the same row are comparable but not necessarily across rows.

### DIFF
--- a/R/type_ridge.R
+++ b/R/type_ridge.R
@@ -10,11 +10,15 @@
 #'
 #' @param scale Numeric. Controls the scaling factor of each plot.
 #' Values greater than 1 means that plots overlap.
-#' @param global.max Logical. Should the height of the individual ridge
-#' densities be scaled relative to the global maximum? Default is `TRUE`.
-#' Changing to `FALSE` will cause a local maximum scaling to be used instead.
-#' Use this latter option if you wish to enforce the same relative height for
-#' each ridge density across `y` categories. 
+#' @param joint.max character indicating how to scale the maximum of the densities:
+#'   The default `"all"` indicates that all densities are scaled jointly relative to
+#'   the same maximum so that the areas of all densities are comparable.
+#'   Alternatively, `"facet"` indicates that the maximum is computed within
+#'   each facet so that the areas of the densities are comparable within each
+#'   facet but not necessarily across facets. Finally, `"by"` indicates that
+#'   each row (in each facet) is scaled separately, so that the areas of the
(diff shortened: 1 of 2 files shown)


zeileis:
P.S.: I noticed that we don't catch the situation where certain groups are too small (1 observation or empty). It would probably be good to handle this more gracefully than throwing an error.

grantmcdermott:
Thanks for this @zeileis. I'm happy to merge, but first want to check:

> P.S.: I noticed that we don't catch the situation where certain groups are too small (1 observation or empty). It would probably be good to handle this more gracefully than throwing an error.

Is this an issue specific to to this PR, or is it a general problem?

zeileis:
I think this was a general problem. Can send an example when the kids are in bed...

zeileis:
It seems that `type_ridgeplot()` generally does not catch empty subgroups, be it empty factor levels directly:

```
iris4 <- transform(iris, Species = factor(as.character(Species), levels = c(levels(Species), "foobar")))
tinyplot(Species ~ Sepal.Width, data = iris4, type = "ridge")
## Error in bw.nrd0(x) : need at least 2 data points
```

Or empty interactions of factor levels:

```
tinyplot(cyl ~ mpg, facet = ~ vs, data = mtcars, type = "ridge")
## Error in bw.nrd0(x) : need at least 2 data points
```

grantmcdermott:
Ah, I see. Do you want to try tackle as part of this PR?

No pressure. I'm happy to this PR merge as-is and address the empty subgroup issue later.

grantmcdermott:
Looks like this is an issue also affecting density types, so let's address it separately.
```r
> tinyplot( ~ mpg, facet = cyl ~ vs, data = mtcars, type = "density")
Error in bw.nrd0(x) : need at least 2 data points
```

PR #296, opened 2025-02-02
Support additional `type_hist()` args

### LINKED ISSUE #293: Additional type_histogram() arguments
Maybe just `freq`, `probability`, and `right` for now.

grantmcdermott:
Closes #293 

(I'm not including `probability` since its duplication with `freq` and is a pain to support with arg passing through our sub function logic.)


grantmcdermott:
Forgot a MWE:

```r
library(tinyplot)
plt(Nile, type = type_hist(freq = FALSE, right = TRUE))
```
tmp

Issue #307, opened 2025-02-07
cex not changing point size for plots with type = "b"


rjknell:
What it says in the title:

plt(runif(10) ~ rnorm(10), type = "b", cex = 2)
plt(runif(10) ~ rnorm(10), type = "b", cex = 1)
plt(runif(10) ~ rnorm(10), type = "b", cex = 0.5)

contrast with:
plot(runif(10) ~ rnorm(10), type = "b", cex = 2)
plot(runif(10) ~ rnorm(10), type = "b", cex = 1)
plot(runif(10) ~ rnorm(10), type = "b", cex = 0.5)

vincentarelbundock:
Thanks a lot for the useful reproducible examples, @rjknell 

I opened a pull request with a fix here: https://github.com/grantmcdermott/tinyplot/pull/308

Waiting for review, and then we can merge.

grantmcdermott:
As an aside, a longer term goal is to enable support for `cex = "by"`, similar to how we support `lty = "by", `pch = "b", etc. This would enable bubble charts and co.

The main blocker is getting two legends to display nicely, which requires some upstream refactoring. (See an older mock-up [here](https://github.com/grantmcdermott/tinyplot/pull/76#issuecomment-1756828564)) But it's on my to-do list.

grantmcdermott:
Closed with #308

PR #308, opened 2025-02-07
Issue #307 `cex` should be respected with `type="b"`

vincentarelbundock:


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: tinyplot
 Type: Package
 Title: Lightweight Extension of the Base R Graphics System
-Version: 0.3.0
+Version: 0.3.0.1
```
grantmcdermott, line 4: Nit: can we stick with the '.99' suffix convention for dev versions? Much as I'd like to increment with each PR, I don't think we'll be disciplined to stick to it and this approach avoids _that_ kind of ambiguity at least. (Following data.table's lead here...)

vincentarelbundock: do we need the date in DESCRIPTION? I don't think I include it in any of my packages. And that seems like an incrementation annoyance as well.

PR #309, opened 2025-02-07
rename test files with type prefix

vincentarelbundock:
We are starting to have a lot of test files. I think it's useful to add a `type` prefix to the type-specific tests. Just helps us navigate and organize the tests better.


grantmcdermott:
Hmmm, looks like an error with one of the `tinyplot_add` tests. Is your fork up to date with main?

(On that note, I've been meaning to say that you should feel free to branch directly off of the main repo for your PRs. Makes it easier for me to test and contribute patches locally. No biggie, though.)

vincentarelbundock:
> (On that note, I've been meaning to say that you should feel free to branch directly off of the main repo for your PRs. Makes it easier for me to test and contribute patches locally. No biggie, though.)

will try to think about it

> Hmmm, looks like an error with one of the `tinyplot_add` tests. Is your fork up to date with main?

There was no see for a jitter snapshot, so I think we were relying on the order of operations and an earlier set seed. So this was a side effects issue.

I set a seed and created a new snapshot. Should be unobjectionable, I think.

grantmcdermott:
> There was no see for a jitter snapshot, so I think we were relying on the order of operations and an earlier set seed. So this was a side effects issue.

Good catch!

PR #314, opened 2025-02-13
Support for barplots

### LINKED ISSUE #305: Wishlist: Support for barplots
For single-sided formulas `~ x` with numeric `x` we already have `type_histogram()` and `type_density()`. But for categorical `x` we are lacking `type_barplot()`. The idea would be that

```
tinyplot(~ cyl, data = mtcars, type = "barplot")
```

would give us essentially the same as

```
barplot(xtabs(~ cyl, data = mtcars))
```

For `~ factor(cyl)` this might even become the default. With a `by` variable we can optionally stack the bars (default, probably)  or put them side by side, i.e.,

```
tinyplot(~ cyl | vs, data = mtcars, type = "barplot")
tinyplot(~ cyl | vs, data = mtcars, type = "barplot", beside = TRUE)
```

should give

```
barplot(xtabs(~ vs + cyl, data = mtcars))
barplot(xtabs(~ vs + cyl, data = mtcars), beside = TRUE)
```

Image

zeileis:
Fixes #305 

This is a first draft of a `type_barplot()` function. It already works quite well but some further improvements are necessary. Also, it would help a lot if `tinyplot()` would be improved to give more control to `type_barplot()` and other `type_` functions.

Basic examples of frequency tables (without `y` variable):

```
tinyplot(~ cyl, data = mtcars, type = type_barplot())
tinyplot(~ cyl | vs, data = mtcars, type = type_barplot())
tinyplot(~ cyl | vs, data = mtcars, type = type_barplot(beside = TRUE))
```

type_barplot

Example for numeric `y` aggregate by `x` (default: `FUN = mean`):

```
tinyplot(extra ~ ID | group, facet = "by", data = sleep, type = type_barplot(beside = TRUE))
```

tinyplot-barplot1

Fancy frequency table:

```
tinytheme("clean2")
tinyplot(Freq ~ Sex | Survived, facet = ~ Class, data = as.data.frame(Titanic), type = type_barplot(),
  facet.args = list(nrow = 1), flip = TRUE)
```

tinyplot-barplot2


zeileis:
**Note:** Currently, you need to `source("type_barplot.R")` because I haven't started the documentation, yet, and hence didn't dare to export the functions.

All of the above could be accomplished without interfering with `tinyplot()` or any of the other functions in the package. However, the current implementation is not idea and it would be good if `type_` functions would have more control over this. In particular:

- **Control over the legend:**  
  Currently, a lot of legend arguments are set automagically inside `draw_legend()` and it is hard to control this from within the `type_`. For the `type_barplot` I faked the type `name = "ridge"` so that the legend comes out ok. However, it would be even better if the levels in the legend were reversed for stacked bar plot. See:  
  https://github.com/grantmcdermott/tinyplot/blob/barplot/R/type_barplot.R#L9
- **Handling of data without y variable:**
  In `tinyplot()` the case without `y` variable is handled by turning `x` into `y` and replacing `x` with an index `1:n`. While this makes sense for the basic single-letter plot types, I think it would be cleaner to leave the data as specified in the formula for custom types. Otherwise the `type_data` has do undo this. See:
  https://github.com/grantmcdermott/tinyplot/blob/barplot/R/type_barplot.R#L20-L23
- **No loop over by levels:**  
  It might be more convenient and/or efficient if there was only a loop over the `facet` levels but not over the `by` levels. Currently, it is cumbersome to get the full data within a facet because `split_data` is split by `by` rather than `facet`. And `data_facet` only contains a single `by` level. Hence, `draw_barplot()` currently binds the `split_data` and splits them subsequently again. See:  
  https://github.com/grantmcdermott/tinyplot/blob/barplot/R/type_barplot.R#L98-L104
- **Access to by aesthetics:**  
  If the `type_draw` wants to draw all `by` levels simultaneously, it needs access not only to the `col`/`bg`/`lty`/`lwd` aesthetics of the current `by` level but of all levels. As this is not passed to the `type_draw` and has not yet been setup in `type_data` the `type_barplot()` has to repeat this code. See:  
  https://github.com/grantmcdermott/tinyplot/blob/barplot/R/type_barplot.R#L58-L67
- **Lack of access to lwd:**  
  In `type_data` we have access to `col`/`bg`/`lty` but not to `lwd`. See:  
  https://github.com/grantmcdermott/tinyplot/blob/barplot/R/type_barplot.R#L68

P.S.: I know Grant is away and Vincent does not have time at the moment. That's perfectly fine, I have to take care of other things as well. I just wanted to write up all of this while I still remember the details.

grantmcdermott:
This is great, thanks @zeileis.

It will probably take me a week or three to review your code, especially since you raise a number of valid design issues in https://github.com/grantmcdermott/tinyplot/pull/314#issuecomment-2655375633 that obviously extend beyond the barplot case. (Not to mention all of the "real work" that's been piling up while I was on vacation.) I might take a crack at "inverting" the nested loop drawing order (to first facets and then by groups) to see what breaks.

zeileis:
As I wrote above, that's perfectly fine!

One small addition, I wondered about: For shading the bars I took the same approach as in `type_boxplot()`, using semi-transparency. However, I wonder whether it would be better to use light opaque colors in both of these cases. In the presence of grid lines, the semi-transparency is not ideal, I think.

zeileis:
I have merged now the "loop over facets first" changes from #331 into this branch and modified `type_barplot()` correspondingly. Getting all data in the current facet is now much easier than before. It still needs a `do.call("rbind", ...)` on the split-up data but that's it. For now this should be good enough, I think. So the third point from my itemized list above has now been addressed. Thanks!

grantmcdermott (commented):
Thanks @zeileis, I finally got some time to look at this properly.

I took a crack at resolving some more of the outstanding issues that you identified. I say "resolve", although some solutions still require automagic steps like special logic in `draw_legend` for `type_barplot()`.

**Aside:** I definitely agree with the general point that we want to transfer as much of legend logic over to the individual `type_*` constructors. But that will require overhauling parts `draw_legend` works for _all_ types, so I think we should take that as a separate issue, rather than implement a unique fix for barplots here.

- Perhaps the easiest way to ensure legend behaviour inheritance via the `type_*` constructors is by allowing them to pass/understand the internal `legend_args` list argument (and then obviously adjust the relevent elements `legend_args` elements like `pt.cex`  as needed). Again, I'll raise this idea in a separate issue.

The one remaining issue that you listed is "**Access to by aesthetics:**". This is an interesting idea and I wonder whether we couldn't (shouldn't) add the levels of each aesthetic as an attribute to said aesthetic? In particularly, when we extract the relevant i<aesthetic> variables here, https://github.com/grantmcdermott/tinyplot/blob/4b3990ea223e8ea4b9ec712058f2df43ad213b86/R/tinyplot.R#L1190-L1194
we could quickly add the original levels  as attributes with:
```r
levels(icol) = col
levels(ibg) = bg
# etc
```

However... before whether we decide on this levels attribute implementation, I'd like to discuss the inconsistency w.r.t. "by" grouping in the current `type_barplot` formulation vs other plot types. (Specifically, the fact that  barplot groups here are drawn at once, rather than separately per  `ii` loop.) Your trick of conditioning on `iby==1` [here](https://github.com/grantmcdermott/tinyplot/blob/d48850503da9f35555b86466df28a4beec119a87/R/type_barplot.R#L98) is clever, but I do worry about introducing a separate logic. (It also results in some data redundancy, but that's relatively minor.)

I see three possible outcomes, in ascending order of complexity:

- leave the current exception for barplots as-is.
- switch to the same general "by" looping iteration as other plot types (i.e., each group drawn separately per loop), perhaps with some logic for correct spacing of the bars. (I know we did something like that for spineplot, but fwiw [this](https://github.com/grantmcdermott/tinyplot/blob/aca8c7bcdfe949e71c88f08596f827ddea59a05d/R/type_boxplot.R#L61-L70) what I did for boxplot types).
- keep the exception for barplots, but switch to a more general logic/mechanism that would allow groups to be drawn once per facet for _other_ plot types too. (Perhaps a logic flag that gets checked before the `ii` "by" loop.) The benefit of this more general approach is that we could leverage color vectorisation for types that support it (e.g., points).

Thoughts?

zeileis:
Grant, thanks a lot for your work on this! A couple of comments:

- Opening a separate discussion for the `draw_legend()` customization in #333 is a good idea. Hence the current workaround is ok.
- The handling of single-sided formulas without `y` variable in `tinyplot.default` is still confusing to me. Are there enough `type_*` functions for which switching from `~ x` to `x ~ index` makes sense? I can understand that for the single-letter types we want to make the switch but I'm not sure beyond that.
- The `levels` attribute trick looks "ok" to me. It's not particularly elegant but it should still be easy enough to understand. If we adopt this I would probably do just one assignment via `icol = structure(col[ii], levels = col)` etc.
- On the question whether plotting all `by` groups in a `facet` should be possible: Personally, I think that we should support this option for efficiency reasons. And if we don't support it, at least we shouldn't actively try to force the groups to be drawn seaparately.
- You picked `ylab = "Count"` as the default y-axis label now (if there is no `y` variable). This makes sense but differs from the default behavior of, say, `barplot(table(mtcars$cyl))`. As we slightly deviate from base R anyway, I don't have strong feelings about this, but just wanted to point out why I had used `""` before, in case you find this relevant.
- One thing I didn't mention so far: Should we avoid repeating the y-axis (or at least the y-axis label) when we use facets with the same y-axis?

grantmcdermott:
Great. Quick replies:

> * The handling of single-sided formulas without `y` variable in `tinyplot.default` is still confusing to me. Are there enough `type_*` functions for which switching from `~ x` to `x ~ index` makes sense? I can understand that for the single-letter types we want to make the switch but I'm not sure beyond that.

I haven't checked in depth, but I take your point. (Our `type_lines()` and `type_points()` equivalents might be additional cases.) Not to kick another can down the road, but let's save this for a separate issue since making an upstream change here would affect multiple other plot types.  

> * The `levels` attribute trick looks "ok" to me. It's not particularly elegant but it should still be easy enough to understand. If we adopt this I would probably do just one assignment via `icol = structure(col[ii], levels = col)` etc.
> * On the question whether plotting all `by` groups in a `facet` should be possible: Personally, I think that we should support this option for efficiency reasons. And if we don't support it, at least we shouldn't actively try to force the groups to be drawn seaparately.

Cool. I'll let you make the call for this particular case then and we can revisit for a wider set of group "vectorisation candidate" types.

> * You picked `ylab = "Count"` as the default y-axis label now (if there is no `y` variable). This makes sense but differs from the default behavior of, say, `barplot(table(mtcars$cyl))`. As we slightly deviate from base R anyway, I don't have strong feelings about this, but just wanted to point out why I had used `""` before, in case you find this relevant.

Thanks, I guessed as much. I think a deviation from base here is an improvement so let's keep the "Count" default.

> * One thing I didn't mention so far: Should we avoid repeating the y-axis (or at least the y-axis label) when we use facets with the same y-axis?

Ideally, yes. But (again) I think this is a general `tinyAxis` issue that needs to be address for several plot types, so that the faceting behaviour follows that of other plot types, i.e., repeated y-axis labels are turned off for non-framed plots. (IIRC we had a similar conversation about spineplot types a little while ago.)

Do you want to add the documentation (assuming you have time) and then we can look at merging this?

zeileis:
OK, I've added some documentation and examples (based on the code above in this PR).

Moreover, I make sure that for `facet_by` we only draw those bars which belong in the current facet. (Before there were zero-height bars included for all other facets.) I also added a `dropzero = FALSE` argument that optionally drops all zeros prior to drawing so that not even the border line of the zero-height bars is drawn.

I also includes `xaxs`/`yaxs` in the `flip` logic (in `tinyplot.default`) so that these are reversed correctly if `flip = TRUE`.

I didn't dare touching the `by` aesthetics stuff because that seemed to be more complex. I hope that this is easier for you?

One final comment: I'm not sure whether drawing semi-transparent bars is really the best way to go. For `type_density` where curves can overlay, the semi-transparent default makes sense. For the barplot my feeling is that it would be more consistent to draw opaque bars by default, using the unmodified colors from the palette. What do you think? (Then we would only need to think about how to use gray instead of black by default.)

zeileis:
Free facet axes currently do not work with `type_barplot()` whereas they do work with other plots:

```
tinyplot(~ cyl, facet = ~ vs, data = mtcars, type = "histogram", facet.args = list(free = TRUE)) ## works
tinyplot(~ cyl, facet = ~ vs, data = mtcars, type = "barplot",   facet.args = list(free = TRUE)) ## fails
```

A number of warnings and errors are created by this:

```
Error in Summary.factor(c(1L, 2L, 2L), na.rm = TRUE) : 
  'range' not meaningful for factors
In addition: Warning messages:
1: In split.default(c(x, xmin, xmax), facet) :
  data length is not a multiple of split variable
2: In split.default(c(y, ymin, ymax), facet) :
  data length is not a multiple of split variable
```

The reason appears to be that `draw_facet_window()` also wants to free the x-axis and not just the y-axis. And because the x-axis is create in a non-standard way (due to the x-variable being categorical) the standard logic for this fails.

grantmcdermott:
I plan to look at this today. Just I've quick note: 

> I also added a dropzero = FALSE argument that optionally drops all zeros prior to drawing so that not even the border line of the zero-height bars is drawn.

I think this should be `drop.zeros` to be consistent with the equivalent arg for `type_histogram`. Also, we set it the default for the latter to TRUE and perhaps we should do the same here. Lmk if you agree and I'll make the change on my end.

zeileis:
Choosing the argument name consistently is certainly a good idea!

As for the value of that default argument: My impression was that it visually works a bit better to draw zero-height bars. If you run the examples of `?type_barplot` there are a couple of illustrations. I think that the visual difference to the histogram is that I have an axis whereas in the barplot I have only a label - and then there is a labeled gap without any bar/line which looks awkward.

But it's not a strong preference. If you feel that not drawing the zero-height bars at all works better visually and/or consistency is more important, just go ahead and change the default value along with the argument name.

grantmcdermott:
Sounds good. I'll leave the zero-height bars for now. I've done some refactoring, which I'll push shortly, and will explain tomorrow (it's late here). But I think we're getting closer to plugging the remaining holes.

grantmcdermott:
Okay, I think this is ready to merge now!

Some final comments:

- I moved all of the data construction steps to  `data_barplot()` instead of `draw_barplot()`. I think this better preserves the idiomatic tinyplot separation logic, and it also simplified some fixes that I need to make w.r.t. overlapping axes in faceted plots, etc.
- Speaking of `draw_barplot`, I ended up simplifying this by just re-using `draw_rect()` instead. I did this after deciding to drop the "vectorised" group drawing for now, and rather stick with the same looping logic that we use for all the other plot types. We can review this again at some point in the future as part of a (potential) generalised vectorised refactor (for the plot types that would support this). But for now I want to avoid special treatment of single plot types.
- We now use solid fill by default (with special "grey" fill if no groups are present). Transparent fill can be enabled in the standard manner, e.g. `tinyplot(~ cyl | vs, data = mtcars, type = "barplot", beside = TRUE, fill = 0.2)`.
- I also fixed the other issues that you mentioned, e.g. "free" facets now work:

``` r
pkgload::load_all(("~/Documents/Projects/tinyplot"))
#> ℹ Loading tinyplot

tinyplot(~ cyl, facet = ~ vs, data = mtcars, type = "barplot",
  facet.args = list(free = TRUE)) 
```



<sup>Created on 2025-03-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

zeileis:
:tada: Fantastic, thanks for resolving all open issues here! And as discussed above, we can revisit some of the more general issues later on. :partying_face:

Issue #322, opened 2025-02-18
Mistake in `type_sanitize`?


grantmcdermott:
In the below code chunk, shouldn't `"lines"` be `"l"`? And the same for `"points"` -> `"p"`?

https://github.com/grantmcdermott/tinyplot/blob/9a567855867e6485b84a10e36b6a893393223891/R/sanitize.R#L49-L53

@vincentarelbundock I believe section goes back to you in https://github.com/grantmcdermott/tinyplot/commit/ec4962b70ce9b4bc73ab5cf0beb72856127563cb#diff-d6fbdd17c7d31c58cbbb80fa0e5be1b555404a61b03ad35e99d6e8cfecec1523

Am I missing something obvious about why these should be "points" and "lines" rather than "l" and "p"?

vincentarelbundock:
I don't think this matters because the defaults for "l" are the same as the defaults for `type_line()`. We could add all the single letters as new lines in the `switch` for completeness, but I don't expect it to have any effect at all, so it's a bit of a waste.

We do need all non-Base string types there, though, so it's important to keep things like "jitter"

grantmcdermott:
> We could add all the single letters as new lines in the `switch` for completeness, but I don't expect it to have any effect at all, so it's a bit of a waste.

I think I might have caught a case where it does matter in https://github.com/grantmcdermott/tinyplot/pull/323#issue-2859150831 (see the final non-plot). I have to run to dinner now, but can follow up later.

grantmcdermott:
Closed with #323

PR #323, opened 2025-02-18
Characters variables inherit same default behaviour as factors

### LINKED ISSUE #321: Default types should be smarter about characters
For example, this should default to `type_boxplot()`.

``` r
library(tinyplot)

set.seed(0)
x <- rep(c("A", "B", "C"), each = 20)
y <- runif(60)
gp <- sample(c("X", "Y", "Z"), size = 60, replace = TRUE)
df <- data.frame(x, y, gp)

plt(
  y ~ x | gp,
  data = df
)
#> Error in Summary.factor(structure(c(1L, 1L, 1L, 1L, 1L, 1L, 1L, 1L, 1L, : 'range' not meaningful for factors
```

Separately, we shouldn't run into errors when `x` or `y` are characters/factors for points anyway. (This is the error above.) I'll address both issues in a PR shortly.

P.S. We coerce characters to factors lower in the main `tinyplot.R` [code](https://github.com/grantmcdermott/tinyplot/blob/cfd5059e4e8e3213ed2425831a00fb26bb0c4c39/R/tinyplot.R#L687-L690), but type sanitizing needs to happen earlier [here](https://github.com/grantmcdermott/tinyplot/blob/cfd5059e4e8e3213ed2425831a00fb26bb0c4c39/R/tinyplot.R#L603).

grantmcdermott:
Closes #321 

``` r
#> ℹ Loading tinyplot

set.seed(0)
x <- rep(c("A", "B", "C"), each = 20)
y <- runif(60)
gp <- sample(c("X", "Y", "Z"), size = 60, replace = TRUE)
df <- data.frame(x, y, gp)

plt(
  y ~ x | gp,
  data = df
)
```

I also addressed/enabled factor variable support for `type_points()`.

``` r
plt(
  y ~ x | gp, pch = 0:2,
  data = df,
  type = type_points()
)
```

Note: the basic `"p"` type isn't working (I think b/c of #322).

``` r
plt(
  y ~ x | gp, pch = 0:2,
  data = df,
  type = "p"
)
#> Error in Summary.factor(structure(c(1L, 1L, 1L, 1L, 1L, 1L, 1L, 1L, 1L, : 'range' not meaningful for factors
```

<sup>Created on 2025-02-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,8 @@ Bugs fixes:
   in #319 and fix in #320 by @kscott-1.
 - Passing `x` and/or `y` as character variables now triggers the same default
   plot type behaviour as factors, e.g. boxplots. (#323 @grantmcdermott)
+- Scatter plots (`type_points()`/`"p"`) now work even if `x` or `y` is a factor
+  or character variable. (#323 @grantmcdermott)
 
 ## 0.3.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,8 +16,11 @@ Bugs fixes:
 
 - The `cex` argument should be respected when using `type="bg"`. Thanks to
   @rjknell for report #307 and @vincentarelbundock for the fix.
-
-- The `lwd` argument should be passed down to `pt.lwd` for type "p". Sets proper line weight for the border of pch symbols in legend. Report in #319 and fix in #320 by @kscott-1.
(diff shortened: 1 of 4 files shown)


grantmcdermott:
(Will add NEWS etc. later tonight. Dinner time first.)

PR #331, opened 2025-03-03
Switch nested drawing loop: First facets, then by groups

grantmcdermott:
_Motivated by various discussions, e.g. https://github.com/grantmcdermott/tinyplot/pull/233#issuecomment-2408778373, https://github.com/grantmcdermott/tinyplot/issues/305#issuecomment-2646688729, and https://github.com/grantmcdermott/tinyplot/pull/314#issuecomment-2661779442._

This PR switches the nested loop that we use for drawing interior plot elements. Specifically, the proposed new nesting logic is:

1. Outer loop over facets
2. Inner loop over groups (i.e., `by` within each facet)

The old nesting logic inverted this relationship&mdash;groups first, then facets&mdash;but this was mostly just an artifact of historical inertia. (I only implemented facet support much later than group support, which was enabled from the get-go.) In addition, the whole datapoints + dedicated types refactoring (#222) has now made it much easier to keep track of the final state of transformed variables (e.g., `by` variables that are created internally as part of the relevant `type_draw` function). All told, looping over facets first seems the most natural way to do things now.

All tests are passing locally for me, but @zeileis and @vincentarelbundock  please take a look if you can. Achim, I should also flag that this will possibly (probably?) break your PR in #314, but hopefully facets first will simplify other parts of `type_barplot()` support. It may be possible to simplify parts of `type_spineplot` and `type_ridge` too, as well as any other types that rely on `tinyAxis` internally, but I didn't look into this in great detail.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -28,6 +28,15 @@ Bugs fixes:
   or character variable. (#323 @grantmcdermott)
 - The `col` argument now accepts a numeric index. (#330 @grantmcdermott)
 
+Internals:
+
+- The order of the nested loop for drawing interior plot elements has been
+  switched. We now loop over facets first (outer loop) before looping over
+  groups second (inner loop), rather than vice versa. The old/inverted nesting
+  logic was mostly an artifact of development inertia and this new nesting logic
+  should simplify the creation of certain plot types. (#331 @grantmcdermott)
+
+
 Misc:
 - Improved column spacing of Arguments in the References section of the website.
   (#328 thanks to @etiennebacher's upstream `altdoc` fix)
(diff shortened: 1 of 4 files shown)


vincentarelbundock:
sorry, probably won't be able to look at this, but if the tests pass, I can't think of a problem

grantmcdermott:
Okay, I'll merge tomorrow. @zeileis lmk if you'd like me to hold off for a bit first, although we can always revert if something goes awry.

zeileis:
Grant, this is great, thank you so much for making this effort!

Indeed it broke the code in #314 but this was easy to fix: I just had to replace `data_by` with `data_facet`. Recombining the split-up `data_facet` is now also much easier than the previous code based on `data_by`, thanks!

PR #340, opened 2025-03-16
Types vignette: list tinyplot types in neater table format

grantmcdermott:
Purely aesthetic website change from the existing bullet points (which feel a bit crowded to me).

Thoughts?  We can merge if other folks think that it makes for an improved reading experience.

Screenshot:

types-tables

### DIFF
--- a/vignettes/custom.css
+++ b/vignettes/custom.css
@@ -1,0 +1,4 @@
+table th {
+    font-weight: normal;
+    font-style: italic;
+}
--- a/vignettes/types.qmd
+++ b/vignettes/types.qmd
@@ -1,5 +1,9 @@
 ---
 title: "Plot types"
+format:
+  html:
+    css: custom.css
+tbl-colwidths: [20,20,50,10]
 ---
 
 **tinyplot** is a lightweight extension of R's base plotting system, designed to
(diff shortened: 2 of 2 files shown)


zeileis:
Very nice, I like it much better!

Small comment: Given that the descriptions are in sentence style, I would also use sentence case for the header row: String, Function, Description, Docs.

grantmcdermott:
Thanks @zeileis. I actually tried this at first, but it felt a little too formal/distracting for my tastes. (I think b/c we already have quite a bit of nesting i.t.o header hierarchy, so adding another layer gets in the way of the content; if that makes sense?)

I need to check that the cross-references to the individual manual pages are working properly&mdash;I know that Quarto can be a bit finnicky about relative paths&mdash;and then will merge.

grantmcdermott:
Minor aside: I was going to put "barplot" under "Shapes", but switched to "Visualizations" b/c it sometimes does automatic aggregation. LMK if you disagree.

zeileis:
Another layer: I was not suggesting to add a layer - but just to say "String" instead of "string" etc.

Barplot as visualization: I agree, slightly better.

grantmcdermott:
> Another layer: I was not suggesting to add a layer - but just to say "String" instead of "string" etc.

Sorry, my reply was confusing (rushing before work). I get your suggestion, but I tried sentence case at first and it felt a bit too formal/heavy some reason. Maybe I'll feel differently in a couple weeks, but for I'll use my veto option to retain lowercase headings :-)

PR #341, opened 2025-03-17
Dealing with y1/y2 ~ x formulas

### LINKED ISSUE #339: Dealing with formula
Hello. It would be great if tinyplot can deal with more complex formulas because it behaves different from base plot. I think this is on the to-do list. I just wanted to put it here 

mtcars

par(mfrow = c(1,2))
plt(mpg/hp ~ wt, data = mtcars)
plot(mpg/hp ~ wt, data = mtcars)

Image

zeileis:
Fixes #339

After the discussion in the issue, I tried to address this with two improvements:

1. Explicit pointer to `I()` in the warning message.
2. More details in the documentation of the formula argument.

Example for 1:

```
tinyplot(mpg/hp ~ wt, data = mtcars)
## Warning message:
## In tinyplot.formula(mpg/hp ~ wt, data = mtcars) :
##   formula should specify at most one y-variable, using: mpg
## if you want to use arithmetic operators, make sure to wrap them inside I()
```

From the `?tinyplot` documentation for argument `formula`:

> a `formula` that optionally includes grouping variable(s) after a vertical bar, e.g. `y ~ x | z`. One-sided formulae are also permitted, e.g. `~ y | z`. Only a single `y` and `x` variable (if any) must be specified but multiple grouping variables can be included in different ways, e.g. `y ~ x | z1:z2` or `y ~ x | z1 + z2`. (These two representations are treated as equivalent; both are parsed as `interaction(z1, z2)` internally.) If arithmetic operators are used for transforming variables, they should be wrapped in `I()`, e.g., `I(y1/y2) ~ x`. Note that the `formula` and `x` arguments should not be specified in the same call.

### DIFF
--- a/R/tinyplot.R
+++ b/R/tinyplot.R
@@ -69,11 +69,14 @@
 #'   features globally for all `tinyplot` plots.
 #' @param formula a \code{\link[stats]{formula}} that optionally includes
 #'   grouping variable(s) after a vertical bar, e.g. `y ~ x | z`. One-sided
-#'   formulae are also permitted, e.g. `~ y | z`. Multiple grouping variables
-#'   can be specified in different ways, e.g. `y ~ x | z1:z2` or
+#'   formulae are also permitted, e.g. `~ y | z`. Only a single `y` and `x`
+#'   variable (if any) must be specified but multiple grouping variables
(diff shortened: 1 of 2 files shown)


grantmcdermott (approved):
lgtm!

PR #345, opened 2025-03-19
xaxt processing in type_barplot

### LINKED ISSUE #344: `type_barplot` does not remove labels of x-axis ticks when specifying `axes = FALSE` or `xaxt = "none"`
Hello!
If you specify `axes = FALSE` or `xaxt = “none”` in combination with `type_barplot`, the labels of the x-axis ticks are still present, while they are removed for other types:
```r
library(tinyplot)

aq = transform(
    airquality,
    Month = factor(month.abb[Month], levels = month.abb[5:9]),
    hot = ifelse(Temp >= 75, "hot", "cold"),
    windy = ifelse(Wind >= 15, "windy", "calm")
)
```
```r
tinyplot(
    Temp ~ Day | Month,
    data = aq,
    type = "l",
    axes = FALSE
)
```
Image
```r
tinyplot(
    Temp ~ Day | Month,
    data = aq,
    type = "barplot",
    axes = FALSE
)
```
Image

zeileis:
Fixes #344 

Previously `xaxt = "l"` was hard-coded. Now the factory-fresh default `xaxt = "s"` gets mapped to `xaxt = "l"` but all other `xaxt` types are still processed. Implications:

- Users can never set `xaxt = "s"` because it always gets overruled. But this might be tolerable because `"s"` never looks great anyway and they can emulate it by adding `xaxt = "a"` for example.
- If themes set another default axis type such as `"t"` ticks will now be drawn. Currently, we don't have such themes, though.

In short, I think this is better than before but maybe still not ideal.

### DIFF
--- a/R/type_barplot.R
+++ b/R/type_barplot.R
@@ -57,7 +57,7 @@ type_barplot = function(width = 5/6, beside = FALSE, FUN = NULL, xlevels = NULL,
 
 #' @importFrom stats aggregate
 data_barplot = function(width = 5/6, beside = FALSE, FUN = NULL, xlevels = NULL, drop.zeros = FALSE) {
-    fun = function(datapoints, col, bg, lty, lwd, palette, xlab = NULL, ylab = NULL, xlim = NULL, ylim = ylim, yaxt, axes = TRUE, facet_by, ...) {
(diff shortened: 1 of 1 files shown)


grantmcdermott:
Thanks @zeileis! I think that this is a good compromise solution that should cover 95% (99%?) of use cases.

Issue #329, opened 2025-02-27
tinyplot not working with layout?


devSJR:
I was trying to use tinyplot with layout. However, there appears to be an issue.

**works:** all plots basically are drawn, but not in the matrix layout

```r
layout(matrix(c(1, 1, 1, 1, 2, 2, 3, 3), nrow = 2, ncol = 4, byrow = TRUE ))

tinyplot(Petal.Width ~ Sepal.Width | Species, data = iris)
tinyplot(Sepal.Width ~ Petal.Length | Species, data = iris)
tinyplot(Petal.Width ~ Petal.Length | Species, data = iris)
```

**doesn't work:** all plots basically are drawn in the matrix layout, when the `by` argument is not used (both with by and formula syntax)

```r
layout(matrix(c(1, 1, 1, 1, 2, 2, 3, 3),
  nrow = 2, ncol = 4, byrow = TRUE
))

tinyplot(Petal.Width ~ Sepal.Width, data = iris)
tinyplot(Sepal.Width ~ Petal.Length, data = iris)
tinyplot(Petal.Width ~ Petal.Length, data = iris)
```

devSJR:
Alright. Would it be possible to give a(n) warning/error message in case layout is used? Something like `trace(layout)`?

grantmcdermott:
Thanks @devSJR that's a good suggestion. At a minimum, we could probably do a better job of documenting this incompatibility.

I want to take a another stab at improving `layout` compatibility with the `legend` behavior, though. It feels to me like maybe we could get this to work, even if facets are a no-go. But I'll only be able to take a look at it in a few weeks.

PR #347, opened 2025-03-20
Make it easier to use tinyplot with pipes

mthulin:
Currently, when `tinyplot` is used with pipes, `data = _`  must be passed to the function:

```
airquality |> subset(Month == 5) |> tinyplot(Temp ~ Wind, data = _)
```

This PR adds the method `tinyplot.data.frame`, which lets users run `tinyplot` after a pipe without having to use the `data` argument:

```
airquality |> subset(Month == 5) |> tinyplot(Temp ~ Wind)
```

This should make it easier for pipe users to adopt `tinyplot` into their workflows.

The function will throw an error if no formula is supplied. Running

```
airquality |> subset(Month == 5) |> tinyplot(Wind, Temp)
```
yields
```
Error in tinyplot.data.frame(subset(airquality, Month == 5), Wind, Temp) : 
  If the first argument to tinyplot is a data frame, then the second must be a formula.
```

I added an example in the documentation, wrapped in \dontrun{} to avoid creating a dependency on R>=4.1.


zeileis:
Thanks for proposing this PR, it's a nice trick.

However, I'm not sure whether we want to give away the possibility to have a dedicated `tinyplot.data.frame` method serving a different purpose, e.g., creating a pairs plot of all possible combinations of variables.

Personally, rather than using the `data = _` argument, I personally just name the first argument:

```
airquality |> subset(Month == 5) |> tinyplot(x = Temp ~ Wind)
```

Personally, I find this easy enough to use. But maybe Grant @grantmcdermott and Vincent @vincentarelbundock feel differently.

vincentarelbundock:
When I first saw this I got super excited and was about to post with great enthusiasm.

But Achim's point about optionality is a **very** good one. I can also imagine other plots where maybe `x` is a data frame with multiple series we want to plot in parallel. Maybe there are also representations of maps as data frames? Not sure if those are real use-cases, but until we explore in that direction, it seems smart to hold off on closing the door altogether.

grantmcdermott:
I won't have time to look at this until tonight. But perhaps we could implement a compromise solution that defaults to pair plots if a formula is not provided, otherwise adopts the trick proposed by @mthulin?

We would probably also need to account for the fact that some data.frames (e.g., `PlantGowth` iirc) have formula attributes for methods dispatch.

mthulin:
The naming-the-first-argument approach is nice! However, I guess that most users usually skip the x when working without pipes, so that they don't write:

```
tinyplot(x = Temp ~ Wind, data = subset(airquality, Month == 5))
```

Personally, I'd prefer not to have to use special syntax to make the function work well with pipes. Similarly, I suspect that users coming from the tidyverse (who of course make up non-negligible proportion of R users) would consider it "a trick you have to do", whereas my solution is more in line with how users would expect things to work.

That being said, I think that @grantmcdermott's idea of defaulting to a pairs plot unless a formula is provided is a great idea. I'd be happy to help with this if you choose to go down that route.

zeileis:
Regarding the suggestion of handling the case with and without formula differently: I'm concerned that the arguments like `type` would have to be handled differently between these two cases. For a pairs plot, for example, it would make sense to choose different types for the diagonal and off-diagonal elements. Hence I would recommend to first work out a proper pairs plot and then see whether the formula trick would be compatible with that without creating too much confusion. (I'm not too optimistic...but probably that's just me).

grantmcdermott:
@mthulin I think we'll need a bit more time to think through the potential implications per @zeileis's last comment. Sorry for the slow action on this. I'm juggling some impending deadlines at work, so my tinyplot dev time is currently limited.

mthulin:
Have you given this any more thought? Others may not agree of course, but for me, the one thing that's keeping me from migrating to tinyplot in my introductory R courses is that the syntax isn't as pipe-friendly as it could be; we use pipes a lot in my courses, and having to add `x = Temp ~ Wind` or `data = _` just makes things confusing for beginners.

vincentarelbundock:
TBC, I'm not the one who's going to make final call on this, but at this stage I think we should hold off on dedicating an important method for this little bit of syntactic convenience.

We all have different teaching philosophies, but I would argue that a pipe-heavy introduction to R should absolutely include a module on argument passing. Otherwise students will try to do things like this:

```r
# breaks
mtcars |> lm(mpg ~ hp)

# works
mtcars |> lm(mpg ~ hp, data = _)
```

So that could just be a nice opportunity to introduce this (very easy) programming concept.

grantmcdermott:
> Have you given this any more thought? 

Sorry @mthulin, not enough to make a call. I think Vincent and Achim have raised enough good points above that we should be cautious about implementing without further testing. FWIW I hope to add a few basic methods to tinyplot in the next release (e.g., `tinyplot.matrix`) and I think we'll be in a better position to evaluate unintended consequences then.

mthulin:
Base R plot() will support this from R 4.6.0: `airquality |> subset(Month == 5) |> plot(Temp ~ Wind)`. So maybe a good time to add a data.frame method with formula support for `tidyplot`?

zeileis:
I'm still not convinced that this is really a good idea. However, if the R Core Team could be convinced and base R is going in this direction, then maybe we should, too. Note that they always had a `plot.data.frame()` already, though, and "just" added the `fomula` handling.

The base R `plot.data.frame()` distinguishes the following three situations if no `formula` is provided:

- If the data has only 1 column: Call `stripchart()` for numeric variables and `plot()` otherwise. (I'm not really sure why stripcharts are used here...it is an awkward default IMO.)
- If the data has 2 columns: Call `plot(x = ..., y = ...)`
- Otherwise: Call `pairs(...)`.

So if we add a `tinyplot.data.frame()` it should do something similar - except for the awkward `stripchart()` I think. A relatively easy setup would be:

```
tinyplot.data.frame <- function (x, formula = NULL, ...) {
  ## original call
  cl <- match.call()

  ## default formula
  ## 1 dim: y ~ 1
  ## 2 dim: y ~ x
  ## 3+ dim: pairs? (not supported, yet)
  if (is.null(formula)) {
    if (ncol(x) > 2L) stop("'formula' is missing with no default for more than 2 columns")
    f = names(x)
    if (length(f) < 2L) f = c("1", f)
    cl$formula = as.formula(paste(f[2L:1L], collapse = " ~ "))
  }

  ## evaluate updated call
  cl[2L:3L] <- cl[3L:2L]
  names(cl)[2L:3L] <- c("", "data")
  cl[[1L]] <- quote(tinyplot::tinyplot)
  eval.parent(cl)
}
```

Thus, 1d and 2d are handled but for 3+ dimensions we haven't got a nice interface yet. But throwing an informative error message would be ok for this setup. Try:

```
cars |> tinyplot()
penguins |> tinyplot(bill_dep ~ bill_len | species, legend = FALSE)
```

tinyplot_df


---

Below I also show a naive implementation of a `tinyplot()` pairs display. The axes take too much space for this to be really feasible, though, that's why I didn't add it to the `tinyplot.data.frame()` function above.

```
par(mfrow = c(7, 7))
for (i in names(penguins)[-8]) for(j in names(penguins)[-8]) tinyplot(penguins[, unique(c(j, i)), drop = FALSE])
```

tinypairs

zeileis:
Grant @grantmcdermott, do you have any thoughts on the version of the `tinyplot.data.frame()` method that I suggested above? Should I create a new PR proposing that method?

grantmcdermott:
Thanks for the nudge. Yes, I agree with your basic outline.

Three questions/thoughts:

1. Should we set a max cap on the number of pair columns (first 7/8 columns only?)
2. It's not a hard blocker, but I guess we need to decide https://github.com/grantmcdermott/tinyplot/issues/598 before a final merge here.
3. It's probably too difficult to shoehorn into this convenience function (let alone the problems that we have with legends and manual `mfrow`/`layout` _a la_ https://github.com/grantmcdermott/tinyplot/issues/329), but it would be cool if we could figure out grouping in a `pairs` display, similar to [this example](https://grantmcdermott.com/quarto-revealjs-clean-demo/template.html#/components-7).

zeileis:
OK, I've set up a separate PR #613 for the proposed function.

All the pairs-style stuff really needs more thought/work. I wouldn't put it into this PR. I've commented on it in the proposed manual page and examples, though.

Also the default single-group color stuff can be done separately. We will have to update many snapshots afterwards anyway.

grantmcdermott:
Superceded by #613

PR #351, opened 2025-03-28
avoid duplication of kernel default arguments

brandmaier:
Avoid duplication of kernel default arguments.

From man page of `match.arg`: 
> In the one-argument form match.arg(arg), the choices are obtained from a default setting for the formal argument arg of the function from which match.arg was called.

### DIFF
--- a/R/type_violin.R
+++ b/R/type_violin.R
@@ -34,7 +34,7 @@ type_violin = function(
         # more args from density here?
         alpha = NULL
     ) {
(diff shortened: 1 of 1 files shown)


zeileis:
Andreas @brandmaier thanks for the suggestion. I'm aware that this works without duplication, however, the lookup in `match.arg()` can be relatively costly. Of course, you may argue that the overhead is small compared to other computations involved. Nevertheless I generally try to avoid it, especially if the possible values are unlikely to change often.

zeileis:
Thanks, Andreas! I had actually never explicitly discussed this aspect with Grant @grantmcdermott - maybe he has a different view on the trade-off :nerd_face:

PR #354, opened 2025-03-31
`type_violin()`

### LINKED ISSUE #348: Violin plots ?
I have a few packages where I now have to add a Suggests: on ggplot2 simply because a demo plots a (micro)benchmark as violin plots via the autoplot feature.  

That would be lighterweight with tinyplot but I don't see violin plots around, or discussed/proposed. Has this come up? Is this feasible especially in the nice autoplot sense of visualizing a benchmarking comparison?

grantmcdermott:
Closes #348 

Quick notes:

- The width of the violins is internally hardcoded at 0.9 units. It would be trivial to add an argument that lets users control this, but what should we call it? `width` would work as an argument name and feels the most natural to me, but would clash with the top-level width argument (which is okay). LMK and I can add an argument.
- I haven't added support for a `joint.max` argument yet. It's almost certainly doable, but I'm very stretched for time and would like to get a "mostly complete" version of violin plots out as part of the dev tinyplot version, if possible.
- I've added some internal catches that enable better support for logged y-axes, but these only "kick in" for the extrapolated densities (i.e., for regions of the densities that extend beyond the original data range). If the original data spans 0 (or negative numbers), then logged axes will trigger an error. (Which feels like the correct approach to me.)
- At some point, it would be nice to support gradient fill for generic polygons (via `draw_polygon`), like we did for `type_ridge`. This seems doable and would enable upstream types to leverage gradient fills for special `y` == `by` cases, e.g. `plt(y ~ x | y, data = data, type = "violin")`. Right now, attempting this kind of `y` == `by` violin plot triggers an error.

MWE (taken from the help docs):

``` r
#> ℹ Loading tinyplot

# "violin" type convenience string
tinyplot(count ~ spray, data = InsectSprays, type = "violin")
```

``` r
# aside: to match the defaults of `ggplot2::geom_violin()`, use `trim = TRUE`
# and `joint.bw = FALSE`
tinyplot(count ~ spray, data = InsectSprays, type = "violin",
         trim = TRUE, joint.bw = FALSE)
```

``` r
# use flip = TRUE to reorient the axes
tinyplot(count ~ spray, data = InsectSprays, type = "violin", flip = TRUE)
```

``` r
# for flipped plots with long group labels, it's better to use a theme for
# dynamic plot resizing
tinytheme("clean")
tinyplot(weight ~ feed, data = chickwts, type = "violin", flip = TRUE)
```

``` r
# you can group by the x var to add colour (here with the original orientation)
tinyplot(weight ~ feed | feed, data = chickwts, type = "violin", legend = FALSE)
```

``` r
# dodged grouped violin plot example (different dataset)
tinyplot(len ~ dose | supp, data = ToothGrowth, type = "violin", fill = 0.2)
```

``` r
# Trimmed and with logged axes
tinyplot(len ~ dose | supp, data = ToothGrowth, type = "violin", fill = 0.2,
         trim = TRUE, log = "y")
```

``` r
# reset theme
tinytheme()
```

<sup>Created on 2025-03-31 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


grantmcdermott:
> Manually putzing with ylim() (while accounting for the logging) kinda works but is not a great UI.

Sure, but in a case like your example those customized args should be set by the `tinyplot.microbenchmark` method, which I'd imagine would have `log ="y", time = TRUE` all turned on by default (just like the equivalent [autoplot.mcrobenchmark](https://github.com/joshuaulrich/microbenchmark/blob/master/R/autoplot.R) method).

We want the (generic) `type_violin` to be as general as possible.

eddelbuettel:
My example showed a bad plot that already had `log ="y"` and `trim = TRUE`.

`microbenchmark:::autoplot.microbenchmark` does not appear to do more than offer a `ymax` arg.  Maybe that is all.  I don't know.

grantmcdermott:
> `microbenchmark:::autoplot.microbenchmark` does not appear to do more than offer a `ymax` arg. Maybe that is all. I don't know.

It has several additional arguments. But I don't want to push this PR further off-track from the narrow goal of providing basic `type_violin` support.  FWIW, under the fold is a basic `tinyplot.microbenchmark` method. I'll submit something like to Josh's repo once we get this PR merged and the next version of `tinyplot` on CRAN.

<details>
<summary>Example: <code>tinyplot.microbenchmark</code> method</summary>

``` r
library(microbenchmark)
#> ℹ Loading tinyplot

tinyplot.microbenchmark = function(x,
                                   order=NULL,
                                   log=TRUE,
                                   unit=NULL,
                                   y_max = NULL,
                                   flip = TRUE,
                                   trim = TRUE,
                                   joint.bw = FALSE,
                                   ...) {
  y_min <- 0
  
  unit <- microbenchmark:::determine_unit(x, unit)
  x$ntime <- microbenchmark:::convert_to_unit(x, unit)
  if (is.null(y_max)) {
    y_max <- max(x$ntime)
  }
  if (!is.null(order)) {
    s <- summary(x)
    x_colnames <- colnames(s)
    order <- match.arg(order, x_colnames, several.ok=TRUE)
    new_order <- do.call("order", c(s[, order, drop=FALSE], decreasing=TRUE))
    x$expr <- factor(x$expr, levels = levels(x$expr)[new_order])
  }
  
  y_label <- sprintf("Time (%s) for neval = %d",
                     attr(x$ntime, "unit"),
                     nrow(x) / length(levels(x$expr)))
  if (log) {
    y_min <- if (min(x$time) == 0) 1 else min(x$ntime)
    log = "y"
  } else {
    log = NULL
  }
  
  tinyplot(
    ntime ~ expr, data = x, type = "violin",
    ylab = y_label, main = "microbenchmark timings",
    log = log,
    trim = trim,
    flip = TRUE,
    joint.bw = joint.bw,
    ylim = c(y_min, y_max),
    ...
  )
  
}

tm <- microbenchmark(rchisq(100, 0),
                     rchisq(100, 1),
                     rchisq(100, 2),
                     rchisq(100, 3),
                     rchisq(100, 5), times=1000L)

ggplot2::autoplot(tm) + ggplot2::theme_classic()
```




``` r


tinytheme('classic')
plt(tm, fill = "white")
```



``` r
tinytheme()
```

<sup>Created on 2025-03-31 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>
</details>

eddelbuettel:
I think we are talking past each other but that is ok -- my specific issue was with my specific dataset; what you wrapped in a function to PR is what I ran already -- see below from a call of your new function leaving all defaults (and in the session where I had already set a theme)  :man_shrugging: 

Anyway, thanks for adding violin support.  The rest, as they say, is details.

image

zeileis:
As for your notes:

- I agree that `width` would be the natural argument here, even if it means that we have to call `type_violin(width = ...)` explicitly rather than `type = "violin"`. We have the same for `type_barplot()`.
- The `joint.max` would be nice to have for consistency. Is the handling here more difficult than in `type_ridge()` or couldn't we just essentially copy the code? (Haven't looked yet.)
- Gradient fill would also be nice but is IMO less important (and more work).

eddelbuettel:
Achim @zeileis do you have thought on what to do with the bad 'tails' that sometimes happen even when trim is set? Anecdata from the set I computed:

```r
> res 
Unit: microseconds           
           expr      min       lq     mean   median       uq      max neval cld        
    zrnormV1(N)  317.544  596.211  647.405  628.181  648.151  4547.39   100 a    
   zrnormV1b(N)  313.803  600.419  630.425  630.668  651.518  3760.85   100 a  
    zrnormMT(N)  397.466  671.140  842.379  704.636  734.760  3980.78   100 a  
 zrnormLZLLV(N)  501.317  783.563  786.975  819.461  837.770  1352.48   100 a
      zrnorm(N)  518.968  823.650  917.488  842.787  867.514  3793.39   100 a     
   zrnormGSL(N)  732.587 1052.041 1078.730 1067.175 1095.515  4146.75   100 ab 
    zrnormQL(N)  942.937 1211.368 1273.459 1248.224 1290.396  4277.85   100 ab  
    zrnormGl(N) 1584.132 1895.564 2035.431 1924.231 1969.382  6195.51   100  b   
       rnorm(N) 4380.368 4644.885 5509.673 4698.672 4842.099 73947.63   100   c    
>  
```

The 73.9 / 4.7 = 15.7 is annoying.  Hardwiring a good decision rule is tricky...  Should we / could offer a trimming percentage as a pass-through option?

Next up is: shall we do one better than the existing function we mock?  I actually quite like the color example @grantmcdermott had above.  When I copy that as `ntime ~ expr | expr` I end up with a different palettem, irrespective of the theme I used.

Anyway, it's all good.  Agree that it is basically there now.

zeileis:
I agree that a good default is tricky. What to do depends also on how much overlap there is between the outliers in one method and the bulk of results for a different method in the benchmark.

As for the coloring: Personally, I like the gray default but would simply expose all arguments (including the formula) so that users can change this as needed for their preferences.

grantmcdermott:
Thanks for the feedback both. I've added the `width` arg and updated the docs.

Given that everyone seems satisfied that this is "good enough" as-is, I'll go ahead and merge in the interests of time constraints. Two things to look at when we get time:

- `joint.max`
- Similarly, right now I've enforced the same width for all violins. We might want to allow separate scaling so that the width of each individual violin is in proportion to their density max.

PR #361, opened 2025-04-04
Centered barplots

### LINKED ISSUE #360: Centering barplots (for population pyramids)
In a [blog post](https://johnmackintosh.com/blog/2025-04-04-base-r-pyramids/) @johnmackintosh mentioned that he was not able to obtain a population pyramid using `tinyplot` and hence ended up writing his own base R code for this. However, this would be doable with the new `type_barplot()` provided that we support centering the bars.

He is looking for something like this:

population pyramid

zeileis:
Fixes #360

Introduce a new argument `center = FALSE` in `type_barplot()` that supports centering the categories, e.g., for population pyramids or for likert scale items.


grantmcdermott:
This looks great @zeileis.

Could you please add an example and also a NEWS.md item?

Other than than these two minor items, I think the only outstanding question is the one that your originally posed [here](https://github.com/grantmcdermott/tinyplot/issues/360#issuecomment-2778430416):

> Open question: What to do about corresponding axis? Instead having both positive and negative labels, both parts of the axis are often displayed with positive labels. Is this something we might also need elsewhere?

I've sort of been thinking about this axis labeling issue in general. `ggplot2` has nice integration with the `scales` package for customized labeling of axes by passing an appropriate function, e,g. [percent labels](https://www.statology.org/ggplot2-percentage-axis/). It would be cool if we could support this kind of functionality too in `tinyplot` and I think the logical place to implement it is in [tinyAxis.R](https://github.com/grantmcdermott/tinyplot/blob/main/R/tinyAxis.R).

In the present case, for example, you could add the following `abs()` transformation just before `do.call()` call here. https://github.com/grantmcdermott/tinyplot/blob/6e941b3fe0851ff0bf6e257c5ecea463284140d6/R/tinyAxis.R#L20

```r
args$at = axTicks(args$side)
args$labels = abs(args$at)
```

If you agree with this assessment, then maybe we hold off fixing the negative axis labels in the PR and rather tackle it in a separate PR that introduces the ability to customize axis labels with an appropriate function.

zeileis:
Ok, I'll integrate your suggestion regarding the axis.

In terms of examples: I couldn't think of a really nice illustration using the data from the `datasets` package. Maybe you have an idea? Otherwise I was thinking that I could also integrate one or both of the data sets from the issue? The aggregated PISA data is 132 x 4 and the population data is 494 x 4. When saved in .rda format they are both just 4K. Or do we want to avoid putting data into `tinyplot`?

grantmcdermott:
> Ok, I'll integrate your suggestion regarding the axis.

Sorry, I'm saying hold off on the axis for now until I manage to address #362. Then we can add `xaxl = abs`  to the `out` object if the `center` argument is non NULL.

> In terms of examples: I couldn't think of a really nice illustration using the data from the `datasets` package. Maybe you have an idea? Otherwise I was thinking that I could also integrate one or both of the data sets from the issue? The aggregated PISA data is 132 x 4 and the population data is 494 x 4. When saved in .rda format they are both just 4K. Or do we want to avoid putting data into `tinyplot`?

Yeah, I think we hold off on any data objects. Let's just use something simple like mtcars, grouped by am for the example. We can put a more "canonical" pyramid example into `gallery.qmd`, possibly pulling from Rdatasets.

zeileis:
OK, got it!

zeileis:
Added an example based on `HairEyeColor` now and mentioned #360 along with #305 in the `NEWS.md`.

grantmcdermott:
@zeileis I think that this is in good shape now. The (default) absolute axis labels for centered plots is working, following #363. From your example:

``` r
#> ℹ Loading tinyplot

tinytheme("clean2", palette.qualitative = c("black", "sienna", "indianred", "goldenrod"))
hec = as.data.frame(proportions(HairEyeColor, 2:3))
tinyplot(Freq ~ Eye | Hair, facet = ~ Sex, data = hec, type = "barplot",
         center = TRUE, flip = TRUE, facet.args = list(ncol = 1))
```



``` r
tinytheme()
```

<sup>Created on 2025-04-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

zeileis:
Cool, this is really nice! Do we want to add a version with `yaxl = function(x) paste0(abs(100 * x), "%")`?

zeileis:
Maybe even support all character specifications of `yaxl` by adding internal versions `"abs_percent"`, `"abs_comma"`, etc. which additionally take absolute values?

grantmcdermott:
SGTM. I won't be able to look at this today (and maybe not tomorrow either). But please feel free to add these to tinylabel.R if you get a chance and then we can adjust barplot internally here too.

zeileis:
OK, done now in PR #364

PR #364, opened 2025-04-07
Absolute numbers in tinylabel() via abs_ prefix

zeileis:
As discussed in #361 all `tinylabel()` specifications can optionally be used with an `abs_` prefix as well which will internally apply `abs()` before using the actual `format_*()` function.

This is convenient for centered bar plots and illustrated now in the `HairEyeColor` example in `?type_barplot`.


grantmcdermott:
Thanks @zeileis, LGTM.

PR #371, opened 2025-04-10
Fix: Irrelevant 'beside' warning for bar plot without groups

### LINKED ISSUE #370: Irrelevant 'beside' warning for bar plot without groups
``` r
tinyplot::tinyplot(extra ~ ID, data = sleep, type = "barplot")
#> Warning in (function (datapoints, col, bg, lty, lwd, palette, xlab = NULL, :
#> 'beside' must be TRUE if there are negative 'y' values
```

<sup>Created on 2025-04-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

zeileis:
Fixes #370 

Only throw the warning regarding `beside=FALSE` (rather than `TRUE`) in the presence of negative `y` values if there are `by` groups without `facet_by` being enabled.

### DIFF
--- a/R/type_barplot.R
+++ b/R/type_barplot.R
@@ -94,7 +94,7 @@ data_barplot = function(width = 5/6, beside = FALSE, center = FALSE, FUN = NULL,
         if (!is.factor(datapoints$by)) datapoints$by = factor(datapoints$by)
         if (!is.factor(datapoints$facet)) datapoints$facet = factor(datapoints$facet)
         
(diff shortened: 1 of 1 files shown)


grantmcdermott:
Thanks @zeileis. If you don't mind, I'm going to close in favour of #372 (which I had started on my local computer before seeing your PR), since it simplifies the codebase in a few other places too. The specific solution for the `beside` case is the exact same, though.

PR #373, opened 2025-04-11
`xaxl`/`yaxl` symbol shortcuts

grantmcdermott:
This PR would allow users to pass the equivalent symbol as a(n even shorter) shortcut to `xaxl`/`yaxl`.  For example, `"percent"` <=> `"%"`, `"dollar"` <=> `"$"`, etc.

``` r
#> ℹ Loading tinyplot
tinytheme('ipsum')

dat = data.frame(
  x = rnorm(100, 1e3, 50),
  y = rnorm(100, 1e3, 10)
)
plt(y ~ x, dat, xaxl = "€", yaxl = "$")
```

``` r
plt(y ~ I(x/1e4), dat, xaxl = "%", yaxl = ",")
```

``` r
tinytheme()
```

<sup>Created on 2025-04-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Thoughts?

### DIFF
--- a/R/tinylabel.R
+++ b/R/tinylabel.R
@@ -10,6 +10,7 @@
 #'   provided: `"percent"` (`"%"`), `"comma"` (`","`), `"dollar"` (`"$"`),
 #'   `"euro"` (`"€"`), or `"sterling"` (`"£"`).
 #' @examples
+#' \dontrun{
 #' x = 1e4
 #' tinylabel(x, "comma")
 #' tinylabel(x, ",") # same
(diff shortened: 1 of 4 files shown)


vincentarelbundock:
I like this A LOT but think we should remove the long shortcuts to avoid duplication and confusion.

grantmcdermott:
Ha, the Vincent "anti-aliasing" principle strikes again!

In seriousness, though I wouldn't want to replace the longer strings. Personally, I'm comfortable keeping both options. Let's see what Achim says...

zeileis:
I have to confess that I also like the idea with the shortcuts and would keep both versions - the short symbol and the long text version.

If we only keep one of the two, I would keep the long text version (and not the short symbol).

grantmcdermott:
Majority rules ;-)

PR #391, opened 2025-05-09
add log(10) labeller

grantmcdermott:
**EDIT:** Also decided to export `tinylabel` as part of this PR. I don't see a reason why not to and it helps to make the documentation cleaner.

Adds support for a `"log"` convenience keyword as part of the `tinylabel` logic. Example use:

``` r
#> ℹ Loading tinyplot
plt(x = 10^c(10:0), y = 0:10, type = "b", log = "x", xaxl = "log")
```

<sup>Created on 2025-05-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Thoughts?

P.S. Here I assume the default is always log10. Should we delineate the other common base cases too? E.g. "log" (= "ln" = "logb"), vs "log10" vs "log2"? Or is that just over-complicating matters?


zeileis:
Off the top of my head I can't think of a good ecampus for another base here.

PR #394, opened 2025-05-14
recordGraphics redux

### LINKED ISSUE #388: Not all doc examples are being displayed on the website
:-/

Seems to be happening across the board, but the main `tinyplot` page is a case in point: https://grantmcdermott.com/tinyplot/man/tinyplot.html#examples

Image

@etiennebacher sorry for the tag, but you've got a knack for figuring these issues out. An ideas?

TBH I'm not sure when this regression occurred. Maybe as result of wrapping some core drawing functions in `recordGraphics` here https://github.com/grantmcdermott/tinyplot/pull/377?

grantmcdermott:
Fixes #388 

I won't share just how many dead-ends I ended going down to resolve this. But I think I've cracked it now.

_tl;dr_ For reasons that I don't fully understand, we can't wrap the whole `draw_legend()` function in `recordGraphics()` without messing up source calls (which is what #388 was ultimately about.) At the same time, simply wrapping the final `legend`/`gradient_legend` calls didn't work either for the Positron case (per #377). _However_, it turns out that the compromise solution is to wrap the fake legend construction (which we need for calculating the plot offset for outside legend placement), the offset calculation, and the final legend plotting call together... but only these steps.

I operationalised this by creating a new internal `recordGraphics(tinylegend(...), ...)` function.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -53,7 +53,10 @@ where the formatting is also better._
 - Fixed dynamic x-axis margin spacing for perpendicular (vertical) label text,
   i.e. cases where `las = 2` or `las = 3`. (#369 @grantmcdermott)
 - Better integration with the Positron IDE graphics pane. Thanks to @thomasp85
-  for the report and helpful suggestions. (#377 @grantmcdermott)
+  for the report and helpful suggestions. (#377, #394 @grantmcdermott)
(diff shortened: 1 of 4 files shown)


zeileis:
Wow!! :tada: Thanks for all your work and for finally resolving this...

Issue #316, opened 2025-02-13
Flatter sidebar on the website


vincentarelbundock:
The sidebar was hard to navigate because it was nested several levels deep.

I made it much flatter and think the result is better, and that we don't lose much intuition.

I'm just opening this issue to flag the change. Feel free to revert or change the structure if you want. Feel free to close when you've made a decision.

zeileis:
Thanks, Vincent, flattening the sidebar is a good idea! Let me take this opportunity, though, to point out that I feel that altdoc's default sidebar menu is not ideal. I think it would be a good idea to make better use of the navigation bar at the top. 

I find the current menu at https://grantmcdermott.com/tinyplot/ particularly confusing on mobile devices. The most notable menu is the hamburger menu in the top left corner. However, this just contains the GitHub link (left screenshot) and the real content is somewhat hidden in the smaller menu below (right screenshot).

![Image](https://github.com/user-attachments/assets/f41a562b-3198-4330-8441-1025d57647e0)

So what I tried to do in my altdoc pages (see e.g., https://topmodels.R-Forge.R-project.org/betareg/) is to mainly rely on the top navigation bar and only use the sidebar when necessary (e.g., for documentation). It would also be nice to generate an overview page listing all documentation topics somewhere.

But let me say that all of this is second order optimization. The altdoc pages n general are great and so is the tinyplot page. I just wanted to raise these issues in case someone else finds this useful. If not, that's also perfectly fine.

grantmcdermott:
Happy with the change, thanks @vincentarelbundock!

> I find the current menu at https://grantmcdermott.com/tinyplot/ particularly confusing on mobile devices. The most notable menu is the hamburger menu in the top left corner. However, this just contains the GitHub link (left screenshot) and the real content is somewhat hidden in the smaller menu below (right screenshot).

Strong agree, @zeileis. The problem here is the underlying Quarto website default, which unfortunately separates the hamburger and secondary menu for some reason (maybe as a compromise for book-based content?) I'll close this current issue to keep things streamlined... But, if you have time, I'd be interested in porting whatever logic you used for the **betareg** website for **tinyplot** so that we _only_ use the hamburger menu. (If I understand correctly, it should be a simple matter of moving sidebar content to the navbar in `altdoc/quarto_website.yml`?)

zeileis:
[...]

PR #395, opened 2025-05-15
altdoc tweaks: navbar, theme, favicon

zeileis:
Grant, today I played a bit around with altdoc and tweaked the tinyplot page a bit. (After meeting Etienne who is currently visiting Innsbruck.) I tried a few improvements which I included in this PR. Maybe you want to check them out and see what you like and what you don't like. Feel free to reject the entire PR if you prefer the current version, that's perfectly fine as well.

- I added a `custom.scss` for the cosmo theme. This uses the teal, pink, and almost-black colors from the hex sticker. Teal is used for the navbar background, pink for links and highlighting, almost-black for the text.
- The picture on the canvas of the hex sticker is used as the favicon.
- The navigation is moved from the sidebar to the navbar. As mentioned in #316 the sidebar navigation is a bit inconvenient on mobile devices. The sidebar is only kept for navigating inside the reference.


zeileis:
Landing page without sidebar:

Screenshot from landing page without sidebar

Reference page with sidebar:

Screenshot from reference page with sidebar


Hamburger menu on emulated mobile device:

Screenshot from hamburger menu on emulated mobile device

grantmcdermott:
I just rendered and previewed everything locally. I really like the proposed changes; it feels like a clear improvement on top of what we already have.

I'm tempted to merge as-is, but two very minor thoughts first:

- While the highlighted "pink" colour looks good against the white background, it's a little garish for my tastes against the teal background. Is there perhaps a softer option that we could try? This is not a strongly held opinion, and you're the color expert, so feel free to go with your preferred aesthetic.
- One collateral benefit of moving everything to the navbar is that we "recover" a menu level. Along those lines, I can see that you already added a "Main functions" entry in the References section and I like this idea. Similarly, I wonder whether it makes sense to create a "Types" menu entry where can nest "Shapes", "Visualizations", and co. under? (Note: This is what we had originally, but we dropped it to avoid too-deep menu nesting... which may be moot now that the navbar has given us one menu level back.)

grantmcdermott:
I had a sec, so actioned the second bullet point above. We can revisit the first later (maybe), but happy to merge now. These changes look fantastic, thanks @zeileis!

zeileis:
- I'll have a look at a better way to highlight the selected navbar link. Either by using a lighter color or by using a different type of highlighting. But I need some time to play with the CSS here.
- For my personal taste, the value added by the additional hierarchy level doesn't justify the additional click. So I would prefer the flatter menu. (Feel free to disagree, of course.)

PR #400, opened 2025-05-18
Support manual breaks (`x/yaxb`)

### LINKED ISSUE #366: Manual breaks
The `scales` package includes convenient `break_*()` functions to manually specify the breaks on axes.

What is the current recommended way to manually specify breaks for ticks? Would it be easy to add support for functions that compute breaks automatically?

https://scales.r-lib.org/reference/index.html#axis-breaks

grantmcdermott:
Closes #366

Operationalised through the new `xaxb` and `yaxb` args. Should play nicely with axis limits and the panel grid, as well as any formatting adjustments applied through `x/yaxl`.

``` r
#> ℹ Loading tinyplot

plt(
  mpg ~ hp | am, mtcars, facet = "by",
  xaxb = c(80, 160, 240, 320),
  yaxb = c(12, 16, 24, 40),
  grid = TRUE
)
```

``` r

plt(
  count ~ spray,
  data = InsectSprays,
  xaxb = c("A", "C", "F"), xaxl = tolower, 
  yaxb = c(10, 30), yaxl = "$"
)
```

<sup>Created on 2025-05-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


grantmcdermott:
@vincentarelbundock I think that this covers the use case/need that you had in mind. Would you mind reviewing or taking a quick look when you get a chance?

R/lim.R
```diff
@@ -15,6 +15,9 @@ lim_args = function(datapoints, xlim, ylim, type) {
   if (identical(type, "boxplot")) {
     xlim = xlim + c(-0.5, 0.5)
   }
+  
+  if (!is.null(xaxb)) xlim = range(c(xlim, xaxb))
```
vincentarelbundock, line 19: Shold this go inside the `if(){}` above? If `xlim` is supplied explicitly, we probably probably don't want to override it with `xaxb`.

vincentarelbundock:
Thanks, this is *exactly* what I needed. A really important feature!  (sorry I didn't get to it in time).

The only comment I have is about whether we should give `xaxb` or `xlim` priority when both are specified and enter in conflict. My gut says that `xlim` should have priority since it is a `base` setting, but I don't feel super strongly about it.

Rest looks great.

zeileis:
Nice! We also need to incorporate that into `type_spineplot()`. The `xaxb` would be the default for `breaks` unless the latter is specified by the user. The `yaxb` should be taken to indicate which of the y-axis categories should be labeled or not.

R/lim.R
grantmcdermott: Yeah, I'm on the fence with this. I guess making an explicit exception for user-supplied `x/ylim` makes the most sense, though. Will action when I'm back at my laptop.

grantmcdermott:
> Nice! We also need to incorporate that into `type_spineplot()`. The `xaxb` would be the default for `breaks` unless the latter is specified by the user. The `yaxb` should be taken to indicate which of the y-axis categories should be labeled or not.

Good catch. So, internally we would just need to make sure that (say) `xaxb` is transferred to `breaks`. 

Question, though: Should we hide these `type_spineplot()` arguments from the user then and rather enforce that they go through the top-level `x/yaxb` arguments for consistency? Alternatively, if we keep them, which takes precedence if the user species both (say) `xaxb` and `type_spineplot(breaks = ...`)?

grantmcdermott:
> Question, though: Should we hide these `type_spineplot()`-specific arguments from the user then and rather enforce that they go through the top-level `x/yaxb` arguments for consistency? I know it would be a breaking change, but I think we should favour a single (common) top-level approach for doing something if possible. Alternatively, if we keep them, which takes precedence if the user specifies both (say) `xaxb` and `type_spineplot(breaks = ...)`?

Sorry, ignore what I was saying above. I was quickly typing on my phone while I was out and hadn't read your comments properly or thought about it clearly. I now recall that `type_spineplot(breaks = ...)` accepts a single number (analogous to histograms), whereas `xaxb` specifies the exact break points. As such, I agree with your comment @zeileis  that "`xaxb` would be the default for breaks unless the latter is specified by the user."

OTOH,

zeileis:
Like in `hist()`, the `breaks` can be both, either the number of breaks or the location of the breaks. So if `xaxb` is specified and `breaks` is not (i.e., `NULL`), `breaks = xaxb` can be used.

grantmcdermott:
Okay, I think we're good to go. Thanks both for the comments and suggestions ;-)

zeileis:
Thanks!

Issue #332, opened 2025-03-07
Support weights


eleuven:
i keep shaking my head at how bad base `R` is at handling weights in a consistent manner, and it would be nice if `tinyplot` at some point would have a good API for them

weights are came up with `type_lm()` (a stylized example below), but weight support would also be very useful for histograms, densities, ...

take this example

```
df  = data.frame(x = c(1,2,3,4), y = c(2,1,4,4), w = c(200,200,1,1))
atx = (1:400)/100

reg  = lm(y ~ x, df)
p    = predict(reg , data.frame(x=atx), interval = 'confidence')

regw = lm(y ~ x, df, weights = w)
pw   = predict(regw, data.frame(x=atx), interval = 'confidence')

# plot
plot(y ~ x, df, ylim=c(-6,8), type = "n")
polygon(c(rev(atx), atx), c(rev(p[ ,3]) , p[ ,2]) , col = rgb(.8,.8,.8), border = NA)
polygon(c(rev(atx), atx), c(rev(pw[ ,3]), pw[ ,2]), col = rgb(1,0,0,.2), border = NA)
points(y ~ x, df, cex = sqrt(length(w) * w / sum(w)))
abline(lm(y ~ x, df), col="black")
abline(lm(y ~ x, df, weights = w), col="red")
```

which produces the following plot 

![Image](https://github.com/user-attachments/assets/a83f96f0-eaea-4296-a87f-c5e1bcc2967c)

this [patch](https://github.com/eleuven/tinyplot/commit/d0a013e1a4dcc61e5436dd652ceb3177b298abd4) is a quick go at adding some functionality to `tinyplot`, which allows for the following code to produce the same plot as above

```
scale = sqrt(length(df$w) * df$w / sum(df$w))
tinyplot(y ~ x, df, ylim=c(-6,8), cex=scale)
tinyplot_add(y ~ x, df, type = type_lm())
tinyplot_add(y ~ x, df, type = type_lm(weights=df$w), col="red")
```

it is however not robust to `NA`s and does not follow (i think) the design logic of `tinyplot`

it was however not clear to me how to add weights to the `datapoints` object in `tinyplot.R` (the correct way?)

anyway, this is just a naive proof of concept to put weights on the `tinyplot` radar

grantmcdermott:
Yeah, weights have come up a few times, e.g. https://github.com/grantmcdermott/tinyplot/issues/272#issuecomment-2540284491

[...]

PR #403, opened 2025-05-18
Handle xmin/xmax/ymin/ymax along with model.frame

### LINKED ISSUE #272: Pass `ymin` and `ymax` through `formula`
Creating a new issue from this comment by https://github.com/grantmcdermott/tinyplot/issues/224 by @zeileis 

> Would we want to be able to specify ymin and ymax through the formula?

zeileis:
Closes #272 

Include min/max arguments in non-standard evaluation of model frame:

- Adds `xmin`/`xmax`/`ymin`/`ymax` to the arguments of `tinyplot.formula` (rather than passing them through `...`).
- Passes along these arguments to `model.frame()` so that variables `(xmin)`, `(xmax)`, etc. are created in the model frame if present (note the brackets).
- Passes `xmin = mf[["(xmin)"]]` etc. on to `tinyplot.default`

Further variables that should be the same length as the data could be added analogously.

Additionally, the `facet` processing is extended so that `facet = var` is handled like `facet = ~ var` in case `var` does not evaluate successfully. This is to avoid confusion about `ymin = var` working but `facet = var` not working.


zeileis:
**Example:**

```
tinyplot(est ~ term, ymin = lwr, ymax = upr, data = coefs, type = "pointrange")
```

now produces the same plot as

```
with(coefs, tinyplot(term, est, ymin = lwr, ymax = upr, type = "pointrange"))
```

e.g., where

```
mod = lm(mpg ~ wt * factor(am), mtcars)
coefs = data.frame(names(coef(mod)), coef(mod), confint(mod))
colnames(coefs) = c("term", "est", "lwr", "upr")
```

**Breaking change:** Specifications that are not the correct length for `data` do not work anymore in the formula interface, e.g.,

```
tinyplot(est ~ term, ymin = c(0, -10), ymax = 40, data = coefs, type = "pointrange")
```

This currently does not work anymore. We could try to work around this but it would be a bit messy, I think.

zeileis:
**Example:**

```
tinyplot(mpg ~ wt, facet = cyl, data = mtcars)
```

is now handled like

```
tinyplot(mpg ~ wt, facet = ~ cyl, data = mtcars)
```

This is with the exception when `cyl` is actually found. Then it is still processed as before, e.g.,

```
cyl <- mtcars$vs
tinyplot(mpg ~ wt, facet = cyl, data = mtcars)
```

This still uses the `cyl` copy of `mtcars$vs` and not `~ cyl`.

vincentarelbundock:
It's kind of amazing that you were able to make this work with so little code. Nice!

grantmcdermott:
Awesome!

This is really great, thanks @zeileis. I don't mind the breaking change at all TBH and the rest of the code looks good to go to me. The only additional things we would need before merging are some doc updates:

- a NEWS item
- add or switch to the formula + NSE-based syntax in some of help files examples, most obviously `type_errorbar` . https://github.com/grantmcdermott/tinyplot/blob/745fa82572482a86de1d55700a5dda7c17cc3bdf/R/type_errorbar.R#L13-L27
- similarly, update to the coefficient plot example in `introduction.qmd` vignette here. https://github.com/grantmcdermott/tinyplot/blob/745fa82572482a86de1d55700a5dda7c17cc3bdf/vignettes/introduction.qmd#L310-L320
- Maybe update the facet argument documentation, since we explicitly mention the requirement/advantage of a one-side formula, e.g. here. https://github.com/grantmcdermott/tinyplot/blob/745fa82572482a86de1d55700a5dda7c17cc3bdf/vignettes/introduction.qmd#L366-L370

Of these, only the NEWS item is strictly necessary. But I'm going to have limited time for tinyplot work in the coming weeks---hence my agitation to get the v0.5.0 update to CRAN ASAP---so would appreciate you actioning as many of the others as you can or see fit.

zeileis:
All done, except the `facet` explanation. I think I wouldn't state explicitly that `facet = var` works and still recommend `facet = ~ var`. Feel free to add something, though, if you think it should be mentioned.

grantmcdermott:
Brilliant, thanks again @zeileis. I think that this end up being one of those "small" changes that ends up having big cumulative benefits from the user side. Plus, as you say, we can deploy the same logic to `weight` (c.f., #332), `label`, etc.

PR #415, opened 2025-06-01
ridge palette

### LINKED ISSUE #414: Wrong color legend in ridge theme/plot
I just noticed that in the ridge plot in the gallery the wrong color legend is drawn:

ridge plot from gallery

The [code](https://github.com/grantmcdermott/tinyplot/blob/main/vignettes/gallery_figs/gallery_08.R) looks alright to me and it did work correctly in version 0.3.0. My guess is that in the recent color handling improvements #410 this was introduced somehow but I didn't dig deeper, yet.

Grant @grantmcdermott maybe you have an idea where to look? Or should I try to debug?

grantmcdermott:
Fixes #414 

Also update some test snapshots while we're at it, so that they use/test against the default ridge theme palette. (The old tests passed a bespoke palette and that's why we missed this regression during the previous release.)


zeileis:
Good idea. I regularly have students who have trouble using recent versions of R because they are not able to upgrade their macOS system and are stuck with something like Catalina. So I guess that there is a market for 4.0.0-compatible packages :innocent:

PR #422, opened 2025-06-06
Better recycling logic for `abline` and co.

### LINKED ISSUE #421: type_hline does not work in non-by facets
When adding `type_hline()` or `type_vline()` to a plain faceted `tinyplot()` (i.e., not via `by`), then the lines are only added in the first panel:

```
tinyplot(bill_len ~ flipper_len, facet = ~ species, data = penguins)
tinyplot_add(type = type_hline(44))
tinyplot_add(type = type_vline(200))
```

non-by faceted tinyplot with hline/vline not working correctly

Interestingly, the same code works for a `facet = "by"` plot:

```
tinyplot(bill_len ~ flipper_len | species, facet = "by", data = penguins)
tinyplot_add(type = type_hline(44))
tinyplot_add(type = type_vline(200))
```

by-faceted tinyplot with hline/vline working correctly

I didn't try to dig into this, yet. Maybe it's also related to the color recycling from #410 ?

grantmcdermott:
Fixes #421


grantmcdermott:
Hmmm, good catch (and question). I haven't tried yet, but perhaps we can introduce some specific logic that reverts back to the default discrete palette for `abline` etc. when `by` is continuous.

zeileis:
My feeling is that it's more complicated and we need to reconsider the `by` handling for these types more generally. For example, I don't understand the idea behind this:

```
tinyplot(bill_len ~ flipper_len | species, data = penguins, type = type_hline(4:6 * 10), lwd = c(30, 18, 6))
```

tinyplot-line

So the length of the `h` argument seems to be reduced to the number of facets but the color from the palette and the line width remain at the number of levels in the `by` variable.

In the continuous `by` case the length of `lwd` is expected to match the length of the palette (100) and not the length of the levels in `by`.

grantmcdermott:
Thanks @zeileis, these cases should be fixed now. I'm not on my normal linux machine at the moment, but will try to use that for adding test snapshots later. Please call out any other edge cases that you come across!

``` r
#> ℹ Loading tinyplot
tinyplot(bill_len ~ flipper_len, facet = ~ species, data = penguins)
tinyplot_add(type = type_hline(44))
tinyplot_add(type = type_vline(200))
```



``` r

tinyplot(bill_len ~ flipper_len | body_mass, facet = ~ species, data = penguins)
tinyplot_add(type = type_hline(45), lwd = 5)
tinyplot_add(bill_len ~ flipper_len, type = type_lm())
```



``` r

tinyplot(
  bill_len ~ flipper_len | species, data = penguins,
  type = type_hline(4:6 * 10), lwd = c(30, 18, 6)
)
```



<sup>Created on 2025-06-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

grantmcdermott:
These are all good cases to think about @zeileis . I've been playing around with some logic that makes more intuitive sense to me. I have't pushed these changes yet---I can if you'd like to play around with them locally---but what do you think of the following?

_(Aside: For some reason, it looks like out warning about `tyoe_hline` and co. only being viable for `add = TRUE` cases no longer holds. I'm okay with this in principle since I think these plots make sense. But we'll need to update the documentation once we settle on a finalized design / logic flow.)_

### Categorical `by`

I find myself in slight disagreement that all of the cases that you presented should result in a single line. (Assuming that you meant this as a point of _desiderata_, rather than a strict reading of the current documentation.) In particular, I think we should default to a single line, but only if the the user hasn't expressly passed a `lwd`/`lty` vector equal to the number of groups. For your examples, this cashes out as:

``` r
tinyplot(bill_len ~ flipper_len | sex, data = penguins,
  type = type_hline(45), lwd = 10, palette = "classic")
```
(single black line, since  `lwd` and `lty` are singlular or null)



``` r
tinyplot(bill_len ~ flipper_len | sex, facet = ~ species, data = penguins,
  type = type_hline(45), lwd = c(21, 7), palette = "classic")
```
(double + color grouped lines, since  `lwd` = no. of groups)



``` r
tinyplot(bill_len ~ flipper_len | sex, facet = ~ species, data = penguins,
  type = type_hline(c(35, 45, 55)), lwd = c(21, 7), palette = "classic")
```
(ditto, albeit with heights varying across facets due to  `h` = no. of facets)



### Continuous `by`

``` r
tinyplot(bill_len ~ flipper_len | body_mass, data = penguins,
  type = type_hline(seq(35, 55, length.out = 100)), lwd = 10)
```
(hlines matched to continuous color gradient due to match the number of groups; an admittedly special case but maybe useful for expert use)



``` r
tinyplot(bill_len ~ flipper_len | body_mass, facet = ~ species, data = penguins,
  type = type_hline(c(35, 45, 55)), lwd = 10)
```
(single hline per facet, since `h` = no. of facets)



<sup>Created on 2025-06-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Thoughts? Again, LMK if you'd like me to push these changes, so that you can grab them and play around with locally.

grantmcdermott:
@zeileis sorry to tag again, but are you okay if I merge this now?

zeileis:
Please wait. I think the cases where the number of lines is the same as the number of `facet` values or the number of `by` values are clear. I'm not yet convinced about the case where we have both at the same time.

Also I'm not sure, yet, whether the length of the graphical parameters should determine the number of lines. Is this consistent with what we do for other plot types?

(I wanted to follow up on this earlier but we've entered the last few weeks of the semester and currently things are a bit busy...)

zeileis:
I still didn't have time to try out code/examples but I tried to think about the desired behavior. So these are currently my thoughts, feel free to disagree with any aspect of this!

1. The number of lines should be determined by the length of `h` (or `v` or `a`/`b`). Thus, I would prefer if `col`/`lty`/`lwd` would adapt to `h` but not vice versa. I feel that drawing the same line with different graphical features is sufficiently rare to justify keeping the interface simpler.
2. The length of `h` must correspond either to
    - 1
    - number of `by` groups (determined by the palette in the continuous case)
    - number of `facet` groups
    - product of the number of `by` and `facet` groups
3. If the number of `by` and `facet` groups is the same, then match to the `by` groups (because this is associated with different graphical features such as color, line type, etc.)
4. If the length of `h` is either 1 or matched to the number of `facet` groups, then the default color should be used that is also used for points in a scatter plot (which is black by default but taken from the `palette.qualitative` theme, if set).
5. If the length of `h` is either the number of `by` groups or the product of `by` and `facet`, then the default colors should be used from the palette used for `by`.
6. I'm not sure which format to expect for `h` in the `by` and `facet` product case: grouping with `by` or `facet` first?

None of this requires fundamental changes but some streamlining which should then also be reflected in the documentation.

grantmcdermott:
HU that I plan to revisit this once #437 is merged.

grantmcdermott:
Hi @zeileis I've tried to strike a compromise solution by following your proposed logic as closely as possible, but also allowing an "escape hatch" for special cases where a user provides additional aesthetic mappings. I don't know if many other people will use them, but I've found them to be valuable for my own use cases and so wanted to retain them as an option. (Again, it's not the default behavior so you'd have to select into it. Plus, the code logic is quite strict, which I think should avoid undesirable outcomes.)

I'm going to merge, since everything is passing... More importantly, I want to submit a v0.4.2 patch version of tinyplot to CRAN today. I'd like to get these bug fixes out as soon as possible (esp. the Positron ones) and also leave enough runway for a potential v0.5.0 release before my UseR2025 talk next month (with new features).

zeileis:
Great, thanks!

Issue #427, opened 2025-06-17
direct labels


grantmcdermott:
I needed this for a plot at work yesterday.

Not sure if we want to roll this into an explicit `type` (and I'm time constrained ATM regardless), but we could at least add it to the `tips and tricks` vignette. (Caveat: I don't think the RHS margin bump will work with themes.)

``` r
library(tinyplot)
palette("classic")

aq = airquality
aq$Month = factor(month.name[aq$Month], levels = month.name[5:9])

# for labels: subset to final dates for each month 
aq2 = aq[aq$Day == ave(aq$Day, aq$Month, FUN = max), ]

plt(
    Temp ~ Day | Month, data = aq,
    type = "l", legend = FALSE, axes = 'l', grid = TRUE
)
text(
    Temp ~ Day, data = aq2, labels = Month,
    pos = 4, offset = 0.2,
    xpd = NA, col = as.integer(Month)
)
```

![](https://i.imgur.com/Kw6hB6F.png)

The above is okay for shorter labels, but cuts off longer labels. We can bump the RHS margin to fix:

``` r
# bump rhs margin for longer labels
omar = par('mar')
omar[4] = omar[4] + max(strwidth(as.character(aq2$Month)))/2
op = par(mar = omar)

plt(
    Temp ~ Day | Month, data = aq,
    type = "l", legend = FALSE, axes = 'l', grid = TRUE
)
text(
    Temp ~ Day, data = aq2, labels = Month,
    pos = 4, offset = 0.2,
    xpd = NA, col = as.integer(Month)
)
```

![](https://i.imgur.com/T44x7cN.png)

``` r

par(op)
```

<sup>Created on 2025-06-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

grantmcdermott:
Yes, but note that the grid lines aren't being clipped now either. So this approach has some undesirable side effects. I think it would be best if we added `xpd` support directly to `type_text()`. Once that's done, the final "tips" entry might look something as follows:

```r
parms = tinyplot:::theme_clean2
parms$mar[4] = parms$mar[4] + max(strwidth(as.character(aq2$Month)))/2

tinytheme("clean2", mar = parms$mar)
plt(
    Temp ~ Day | Month, aq, type = "l", legend = FALSE, 
    draw = plt(
        Temp ~ Day | Month, 
        data = aq2,
        type = "text", add = TRUE,
        labels = aq2$Month,
[...]

PR #428, opened 2025-06-20
Support `xpd` and `srt` in `type_text()`

grantmcdermott:
Follow up to discussion in #427.


vincentarelbundock:
thanks for this! It's kind of nice to see how easy it is to add arguments like this.

PR #429, opened 2025-06-20
Direct labs tip

### LINKED ISSUE #427: direct labels

grantmcdermott:
Closes #427


grantmcdermott:
I had to implement bit of `echo: false` / `eval: false` trickery make this play nicely with Quarto. (`strwidth()` needs an open device to work, but this isn't how Quarto chunks behave.) Looks good on the rendered website, though.

PR #431, opened 2025-06-24
fix numeric xlevels specification in type_barplot

### LINKED ISSUE #430: Unclear error for invalid `xlevels` param in barplot
### Problem

I wanted to reverse the order of the x axis in a bar chart. I eventually succeeded with this:

```{r}
tinyplot(
  ~cyl,
  data = mtcars,
  type = "barplot",
  beside = TRUE
  ,xlevels = levels(factor(mycars$cyl))[3:1]
)
```

However, the helptext "a character or numeric vector specifying in which order the levels of the x variable should be plotted. " does not make it clear that that each element of the vector must be a match for one of the levels of x, not an index. I would have initially expected `xlevels  = 3:1` (indices) and/or `xlevels = c(8, 6, 4)` (actual x values passed in). These resulted in 

```
Error in (function (datapoints, col, bg, lty, lwd, palette, xlab = NULL,  : 
  object 'x' not found
```

which does not indicate why my values are invalid or what valid values might be.

### Suggested Remedies

1) Fix the error message: run a check on the xlevel param to check if it matches levels of x. If not, error with something like "xlevels must match levels of x".
2) Clarify the parameter description as to what values are acceptable and what they do.
3) Add an example to barplots examples or gallery where xlevel is set.
4) Accept numeric input. Either if x is numeric, allow xlevels to be values of x OR accept level indices. (Obviously doing both would create conflicts, but seems like one or the other could be done.)

Hope this is helpful!

zeileis:
Fixes #430 

- Numeric `xlevels` specifications in `type_barplot()` never worked correctly because the `x` rather than `datapoints$x` was used in one place. Fixed now.
- Improved the documentation of `xlevels` and added two simple examples.
- Added a warning in cases where `xlevels` contains entries that do not correspond to any level of `x`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,7 +16,9 @@ where the formatting is also better._
 - Safer handling of pre-plot hooks. Resolves an issue affecting how `tinyplot`
   behaves inside loops, particularly for themed plots where only the final plot
   was being drawn in Quarto/RMarkdown contexts. Special thanks to @hadley and @cderv
-  for helping us to debug. (@vincentarelbundock #425)
+  for helping us to debug. (#425 @vincentarelbundock)
+- The `xlevels` argument of `type_barplot()` could not handle numeric indexes correctly
+  (#431 @zeileis)
 
 ## 0.4.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,7 +17,7 @@ where the formatting is also better._
   behaves inside loops, particularly for themed plots where only the final plot
   was being drawn in Quarto/RMarkdown contexts. Special thanks to @hadley and @cderv
   for helping us to debug. (#425 @vincentarelbundock)
-- The `xlevels` argument of `type_barplot()` could not handle numeric indexes correctly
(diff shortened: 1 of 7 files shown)


vincentarelbundock:
Thanks for the fix! Maybe insert a test like this:

```r
f = function() {
  tinyplot(
    ~cyl,
    data = mtcars,
    type = "barplot",
    beside = TRUE,
    xlevels = levels(factor(mycars$cyl))[3:1]
  )
}
expect_snapshot_plot(f, label = "barplot_xlevels_issue430")
```

zeileis:
Similar improvements have now been done for `type_spineplot()` as well where `xlevels` was also added (previously we only had `ylevels`).

zeileis:
Great, thanks for the feedback, this is very helpful. Should we wait for feedback from the original poster?

zeileis:
OK, I will let her know when the dev version is on R-universe.

grantmcdermott:
Thanks all, LGTM.

PR #433, opened 2025-06-29
Support bubble plots

### LINKED ISSUE #432: `"bubble"`

grantmcdermott:
Once ready, should fix #432

I've had to impose some constraints to make this work (e.g., the dual legend only prints on the RHS) and there are almost certainly multiple corner cases that I haven't tested against. But the basic functionality seems to be there.

MWEs:

``` r
devtools::load_all("~/Documents/Projects/tinyplot/")
#> ℹ Loading tinyplot
tpar(grid = TRUE)

# discrete by groups & cex = "by"
tinyplot(
  Sepal.Length ~ Petal.Length | Species,
  data = iris,
  cex = "by"
)
```

``` r
# dual legend
## (it's not obious here, but the code is smart enough to adjust spacing,
## depending on the widest one)
zz = iris$Sepal.Length
tinyplot(
  Sepal.Length ~ Petal.Length | Species,
  data = iris,
  cex = zz
)
```

<sup>Created on 2025-06-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


grantmcdermott:
> Is it possible to scale all `cex` values up or down by some constant, e.g., when the plot symbols are too small?

Yup, use the `clim` ("cex limit") argument. The default values are `c(0.5, 2.5)`, but you could try, say, `clim = c(1, 3)`. One caveat: values greater than 2.5 will lead to overlapping circles in the legend. I guess we could fix this by simultaneously adjusting the `y.intersp` parameter, but I haven tried/done this yet.

> Would it be worth adding a separate `type_bubble()` for this? This could get an option to count frequencies and display these in `cex`. That would be useful for discrete data (e.g., counts etc.).

Hmmm, that's an interesting idea and I'm not against it _a priori_. As an aside, I've been wondering whether we should support a `size` argument alias for `cex`, and maybe this would make more sense for a dedicated `type_bubble()` type. (Apart from being more intuitive, would NSE support be simpler for a dedicated `size` argument that only affects one plot type (compared to `cex`)?

zeileis:
I haven't had time to really play around with this. So just three comments:

- _Legend:_ It might be worth checking out whether `ggplot2` is using some special trick for these types of displays.
- _Size alias:_ For `type_bubble()` this might be ok but for `tinyplot()` I would probably not start introducing aliases.
- _NSE:_ I think that NSE would in principle be possible for either `cex` or `size`. Off the top of my head I don't see where this would make a relevant difference.

grantmcdermott:
@zeileis @vincentarelbundock I think that this PR is now ready for review. I think that I've managed to plug the main issues that came up in testing and everything seems to work as expected. Dual legend support required some internal changes that made the code a bit uglier, but I don't think that can be avoided.

Some quick examples:

```r
tinytheme("clean")

plt(
  Life.Exp ~ Income | Illiteracy < 1, data = data.frame(state.x77),
  cex = state.x77[, 'Area'],
  file = "bubble1.png", width = 8, height = 5
)
```
bubble1

In the absence of NSE support for `cex` (see PS below), we can use `with()` instead of `tinyplot.formula`... Slightly fancier version of the previous plot.

```r
with(
    data.frame(state.x77),
    plt(
        x = Income, y = Life.Exp,
        by = Illiteracy < 1,
        pch = 19, alpha = 0.5,
        cex = Area, clim = c(1, 5),
        main = "Life expectancy among US states (1977)",
        file = "bubble2.png", width = 8, height = 5
    )
)
```

bubble2

Dual continuous × continuous legends work too.

```r
with(
    data.frame(state.x77),
    plt(
        x = Income, y = Life.Exp,
        by = Population,
        pch = 21, bg = 0.4,
        cex = Area, clim = c(1, 6),
        main = "Life expectancy among US states (1977)",
        file = "bubble3.png", width = 8, height = 5

    )
)
```

bubble3


P.S. I couldn't quite NSE processing with `tinyplot.formua(..., cex = <variable>)` to work properly. Specifically, I could correctly pass process the cex variable as part of the model frame, and ultimately generate the bubble legend... but the legend title is wrong. @zeileis perhaps we can troubleshoot this issue as part of the current PR or a separate one (as you prefer).

zeileis:
**NSE:**

- Currently, when you use `tinyplot.formula` and specify `cex` then it gets passed through the ellipsis `...` and is not evaluated. Thus, in `tinyplot.default` it is possible to `deparse1(substitute(...))` it as you do in https://github.com/grantmcdermott/tinyplot/blob/bubble/R/tinyplot.R#L761
- However, when you extract `cex` from the model frame in `tinyplot.formula` then you call `tinyplot.default(..., cex = cex, ...)` explicitly. Hence, the deparsing will always pick up `"cex"` in `cex_dep`.
- If we want to preserve the original variable label we need two things: (1) Capture it within `tinyplot.formula`. (2) Pass it on to `tinyplot.default`.
- For (1): We cannot get the original label from the model frame because it is always `"(cex)"` but we can deparse it from the call, e.g., `cexlab = if ("(cex)" %in% names(mf)) deparse1(m$cex) else NULL`.
- For (2): To pass the `cexlab` information to `tinyplot.default` we either need to introduce a new argument (just like `xlab` and `ylab`) or we use an attribute, e.g. `attr(cex, "cex_label") = cexlab`. The former is probably overkill. The latter is somewhat more obscure but similar to what we do with the `facet` attributes.

grantmcdermott:
Thanks folks. Unfortunately, I've uncovered a bug when testing out altdoc rendering locally. It seems that we're re-triggered the infamous "hidden" plot issue from https://github.com/grantmcdermott/tinyplot/issues/388. To be more precise, it only affects dual legends and only for non-interactive calls... but it needs to e fixed. At least, this time  I know why it's happening... even though I haven't had time to think of a solution.

I'm quite pressed for for time with other deadlines ATM (and also need to start in my useR! 2025 slides)  but will update here if/when I get a chance.

> I'm still a bit uneasy about adding all of this directly in the scatter plots, though. Wouldn't a dedicated `type_bubble()` be cleaner? Or would it lead to too much code duplication?

Yeah, the latter and I also think that consistency kind of requires that we support this through cex. But Let' me reevaluate/ after I crack the above issue.

zeileis:
Sure, thanks for the feedback!

Re: Support through cex. This is also passed to the type functions, so in principle it would be possible. But I guess the legend still would require some special handling?

grantmcdermott:
I finally had some time to solve the missing plot issue. I also ended up simplifying/limiting the changes to the main `tinyplot`, by outsourcing the legwork to a new (internal) `draw_multi_legend()` function. 

This is almost ready for final review and discussion. I'd just like to take a crack at the NSE/labeling issue, but that will have to wait until this evening or tomorrow.

grantmcdermott:
_**Note:** This is mostly a note to myself for posterity. Everyone else should feel free to ignore._

One way to create dual legends in vanilla base plot is by using an inset of 0.5 plus an adjustment term. Assuming we don't want a gap between the two legends (i.e., they can be directly adjacent), we have:

- For the 'top' legend this adjustment term is zero, since the 0.5 inset already forces the (bottom of the) legend rectangle relative to the middle of the plot
- For the 'bottom' legend, the adjustment term is the top height of the rectangle, adjusted in terms of normalized plot coordinates (`"npc"`), since we want to push this second legend down by its entire height.

A simple example to illustrate:

``` r
plot(0:10, type = 'n')
abline(h = 5, lty = 2)

legend("bottomright", "a", pch=1, title= "Legend 1",inset = c(0, 0.5))

l2 = legend("bottomright", c("b", 'c'), pch=1, title= "Legend 2", plot = FALSE)
# l2h = l2$rect$h ## use top, rather than h
l2t = l2$rect$top
l2h = grconvertY(l2t, to = 'npc')
legend("bottomright", c("b", 'c'), pch=1, title= "Legend 2",inset = c(0, 0.5-l2h))
```



<sup>Created on 2025-09-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

However... try as I might, I have not been able to make this simple approach work for `tinyplot`. I honestly don't know why; it could be because of `recordGraphics` adjustments, our bespoke margin tweaks, or something else. 

Nevertheless... another approach that works for dual axes is to do the same 0.5 inset, but from opposite edges (e.g., "bottomright" vs "topright"). In this case, the solution is even simpler since you don't need to calculate an additional adjustment term:

``` r
plot(0:10, type = 'n')
abline(h = 5, lty = 2)

legend("bottomright", "a", pch=1, title= "Legend 1",inset = c(0, 0.5))

legend("topright", c("b", 'c'), pch=1, title= "Legend 2", inset = c(0, 0.5))
```



<sup>Created on 2025-09-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The good news is that this latter approach _does_ work with `tinyplot` and this is what my PR here does underneath the hood.

## So what?

The second approach is simpler and gives exactly the same result for this dual legend case. So, why worry? The problem is that it can't scale for multi-legends (> dual legends), b/c it fully depends on the "opposite edges" logic. So if we'd like to support, say, three-part legends down the road, then we'd have to figure out why the first approach isn't working. And then you could do something in spirit of:

``` r
plot(0:10, type = 'n')
abline(h = 5, lty = 2)

l1 = legend("bottomright", "a", pch=1, title= "Legend 1", plot = FALSE)
l2 = legend("bottomright", c("b", 'c'), pch=1, title= "Legend 2", plot = FALSE)
l3 = legend("bottomright", "d", pch=2, title= "Legend 3", plot = FALSE)

l1h = grconvertY(l1$rect$top, to = 'npc')
l2h = grconvertY(l2$rect$top, to = 'npc')
l3h = grconvertY(l3$rect$top, to = 'npc')

lh_tot = l1h + l2h + l3h
adj = 0.5+lh_tot/2

l1 = legend("bottomright", "a", pch=1, title= "Legend 1", inset = c(0, adj-l1h))
l2 = legend("bottomright", c("b", 'c'), pch=1, title= "Legend 2", inset = c(0, adj-l1h-l2h))
l3 = legend("bottomright", "d", pch=2, title= "Legend 3", inset = c(0, adj-l1h-l2h-l3h))
```



<sup>Created on 2025-09-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Until then, we're stuck with dual legends at a maximum.

grantmcdermott:
@zeileis @vincentarelbundock This is now ready for final review.

(After a bit of reflection, I decided not to implement the `cex` NSE support as part of this PR, since it would imply diverging API logic for this graphical parameter versus the others (`lty`, `lwd`, etc.) This is something we can tackle in a separate issue; see also #455.)

I think the only remaining thing is to decided whether we want to introduce a separate `type_bubble()` type, which would be a thin layer over the functionality that I have rolled into `type_points(..., cex = <>)` here? Some quick thoughts:

- I think it's important to retain the `type_points(..., cex = <> functionality that I have introduced here. since that brings equivalent `cex` functionality to what we have long supported for the other graphical parameters, e.g. `cex = "by"` etc.
- If that's correct, then I'm still not entirely sure what the benefit of a separate `type_bubble()` type would be besides mnemonic value. Is it to add an optional aggregating function that would be applied to the data internally?

zeileis:
Quick feedback for the last question: Yes, aggregation would have been the main motivation. Plus potentially introducing additional arguments with mir intuitive names IIRC.

grantmcdermott:
Got it. Okay, let me try to take a stab at that later today. Hopefully it's a simple overlay and then we merge this and get to some of the other PRs/issues.

grantmcdermott:
Ugh, sorry to go back and forth on this, but after a bit of testing I think it's best just to merge this as-is, without trying to introduce the `type_bubble()` overlay. One thing I've run into is the lack of consistency of potential aggregating functions. For example, the `xtabs()` example that you use above takes a one-sided formula, whereas other aggregating functions would take a double-sided formula. This could probably be solved with more time and thought, but I feel we need to move on not and unlock the other PRs. We can always revisit and tweak down the road. The most important thing IMO is that the main bubble plot and dual legend functionality is now working.

PR #437, opened 2025-07-06
Unified `recordGraphics`

### LINKED ISSUE #313: Added elements are misaligned with facets
Can I get more clarification on how to draw separate elements using the facet argument? I share this in case it helps work with implementing additional facet elements:

```
# generate data
lambda1 <- 1.8
lambda2 <- 0.7
x <- 0:10
p1 <- dpois(x,lambda1)
p2 <- dpois(x,lambda2)
dp <- data.frame(x=c(x,x),p=c(p1,p2),lambda=rep(c(lambda1,lambda2),each=length(x)))
dp$lambda <- as.factor(dp$lambda)
# plot
plt(p ~ x,
    facet = ~lambda,
    data=dp,
    type="h",
    lwd=2,
    col="skyblue",
    draw=c(axis(1,at=x,labels=x),
           points(p~x,data=dp,col="black")))
```

As we can see, it draws both sets of points (not divided by facet). I can't add facet to the points command in the draw argument (of course, as its a base R command).

I could only get this to work by using plt_add(), but here I cannot manage shared y-axis limits (if I remove the free=TRUE argument, the points are drawn offset from where they should be).

```
plt(p ~ x,
    facet=~lambda,
    data=dp,
    type="h",
    lwd=2,
    col="skyblue",
    facet.args=list(free=TRUE),
    ylim=c(0.,0.5)
)
plt_add(type="p")
```

Thank you very much!

grantmcdermott:
Fixes #313
Fixes https://github.com/posit-dev/positron/issues/7316

This PR wraps all of the tinyplot drawing logic in a single, unified `recordGraphics({...})` call and thereby resolves the layering + resizing issues that we have been running into for faceted plots, as well as some related problems specific to the Positron IDE. The code is a little (tiny?) bit uglier, but that's a trade-off I'm happy to make if we can avoid unexpected plotting errors when adding layers.

(Previously, we had separate `recordGraphics()` logic for the `draw_legend` and `draw_facet_window` stages, but something was obviously falling through the cracks that was leading to misalignment of added layers.)

My screen (video) recorder isn't working for some reason, but here's a simple screenshot of a correctly aligned layers from a Positron session.

resize-fixed-positron

Please give this a quick test drive when you get a sec @zeileis, @vincentarelbundock and/or @jhpantel. Thanks!

### DIFF
--- a/R/draw_legend.R
+++ b/R/draw_legend.R
@@ -358,7 +358,6 @@ draw_legend = function(
     #
     # Note: We wrap everything in `recordGraphics()` to preserve legend spacing
     # if the plot is resized (also necessary for Positron graphics logic regardless)
-      recordGraphics(
         tinylegend(
           legend_args = legend_args,
           ooma = ooma,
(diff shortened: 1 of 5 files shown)


zeileis:
Thanks, this is great! I've played around with a couple of examples which all work nicely now.

grantmcdermott:
Update: This appears to be a recurrence of #388 (resolved in #394).

What's really odd (annoying) is that the problem is not consistent across IDEs or runtimes.

- If I run `example(tinyplot)` or `source('test.R')` from Positron, then all of the plots render correctly.
- If I try the exact same commands from RStudio, then I run into the same problems as #388.

Obviously, calling `altdoc::render_docs()` suffers from the same issue as RStudio...

zeileis:
I played around with your `test.R` script:

```
library("tinyplot")
tinyplot(Sepal.Length ~ Petal.Length | Species, data = iris)
# Sys.sleep(1e-1)
tinyplot(Sepal.Length ~ Petal.Length | Species, data = iris, pch = 16, cex = 2)
```

Instead of doing the `Sys.sleep(1e-1)` in between the plots, it also works to do `plot.new()` in between the plots. This also has the effect that both plots show up in my RStudio plot history. Maybe it helps to put the `plot.new()` call in a different position within the `recordGraphics()` - or maybe to put it outside the `recordGraphics()`? But I'm really just guessing here...

grantmcdermott:
Thanks Achim. Fingers crossed I've found a simpler workaround. Will test
after dinner.

Grant McDermott
*www.grantmcdermott.com <http://www.grantmcdermott.com>*
+1 805 895 7480

grantmcdermott:
Closing in favour of #438.

PR #439, opened 2025-07-09
Move altdoc to config

grantmcdermott:
Testing the setup suggested by @etiennebacher here: https://github.com/etiennebacher/altdoc/issues/322#issuecomment-2794929124

This is so we can use the dev version of altdoc, but avoid having to comment things out for CRAN submission.

(I'm not sure that it will work given our reliance on Dirk's r-ci script, but let's see what CI throws up.)

### DIFF
--- a/.github/workflows/altdoc.yaml
+++ b/.github/workflows/altdoc.yaml
@@ -34,15 +34,18 @@ jobs:
 
       - name: Dependencies
         run: ./run.sh install_all
+        with:
+          needs: website
+          cache: "always"
 
       - name: Build site
         run: |
           # If parallel = TRUE in render_docs()
-          # future::plan(future::multicore)
+          future::plan(future::multicore)
           install.packages(".", repos = NULL, type = "source")
           install.packages("pkgload")
           pkgload::load_all()
-          altdoc::render_docs(parallel = FALSE, freeze = FALSE)
+          altdoc::render_docs(parallel = TRUE, freeze = FALSE)
         shell: Rscript {0}
 
       - name: Deploy to GitHub pages 🚀
--- a/.github/workflows/altdoc.yaml
+++ b/.github/workflows/altdoc.yaml
@@ -34,9 +34,12 @@ jobs:
 
       - name: Dependencies
         run: ./run.sh install_all
-        with:
-          needs: website
-          cache: "always"
+      
+      - name: Altdoc from GitHub
+        run: ./run.sh install_github etiennebacher/altdoc
+
+      - name: Additional Altdoc Dependency
+        run: ./run.sh install_r future.apply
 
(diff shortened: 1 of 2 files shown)


grantmcdermott:
Hmmm, this somehow seems have removed the altdoc CI workflow completely :-/

eddelbuettel:
I am confused. I have no idea what you are trying to do / what gave you the idea this would be supported. `r-ci` uses a _fixed_ stanza because most normal uses are covered by what is `DESCRIPTION` file.  When it is not, _e.g._ when one needs extras the standard, simple and obvious way is to just add a stanza.  For example, CRAN and R do not have a way to add a system dependency declaration so when `RcppGSL` needs `libgsl-dev` we just install it:

https://github.com/eddelbuettel/rcppgsl/blob/master/.github/workflows/ci.yaml#L29-L32

Likewise when `RProtoBuf` needs a few packages related to Protocol Buffers:

https://github.com/eddelbuettel/rprotobuf/blob/master/.github/workflows/ci.yaml#L32-L35

And so on.  You could wrap that in an action with arguments, and if you think it adds value, have random new fields from `DESCRIPTION` parsed (not something I do).   

At then end of the day a GHA is just three shellscripts pretending to be a yaml trenchcoat.

.github/workflows/altdoc.yaml
```diff
@@ -34,15 +34,18 @@ jobs:
 
       - name: Dependencies
         run: ./run.sh install_all
+        with:
+          needs: website
+          cache: "always"
```
eddelbuettel, line 39: You need the logical equivalent of the declaration of line 55 below added here, potentially in two (OS-dependent) blocks depending on which OS you want to run here.  If it is 'merely' an Linux-only documentation build step you do not need to condition and can just list what is needed as arguments to `apt`, or to `install.r` (wrapping `install.packages()`) if you want it to be R.  Similarly, `installGithub.r` fetches from where you think it does etc.

eddelbuettel:
So I think this may work:

```yaml
      - name: Dependencies
        run: |
            # default packages
            ./run.sh install_all
            # altdoc from GH
            ./run.sh install_github etiennebacher/altdoc
            # plus future.apply
            ./run.sh install_r future.apply
```

[ Untested though. ]

grantmcdermott:
Thanks Dirk, will test after work.

(Also: had a good chuckle at "GHA file is just three shellscripts hiding inside a yaml trenchcoat".)

eddelbuettel:
Tried to help you our with a test via a quick fork-and-pr (in #443) alas no action was triggered.

grantmcdermott:
I integrated your changes directly in this PR and... huzzah, that seems to have done it. Thanks @eddelbuettel!

PR #445, opened 2025-07-15
Revert to remotes

grantmcdermott:
R-universe build doesn't work with r-ci (probably for override reasons) and just reverts to r2u/CRAN. (Details [here](https://github.com/grantmcdermott/tinyplot/actions/runs/16306381497/job/46053266747#step:7:165).) So now we try with `remotes` again and explicit GH token...


eitsupi:
> R-universe build doesn't work with r-ci (probably for override reasons)

I suppose you should disable bspm.
https://cran4linux.github.io/bspm/reference/integration.html

PR #446, opened 2025-07-20
Better horizontal legend spacing and multicolumn support

### LINKED ISSUE #434: Better default spacing for horizontal legends
The default spacing for horizontal legends isn't great IMO, particularly if the keys aren't of the same length. This is because they all take the width of the longest key. Example:

``` r
library(tinyplot)
aq = transform(
  airquality,
  Month = factor(month.name[Month], levels = month.name[5:9])
)

plt(Temp ~ Day | Month, data = aq, legend = "bottom!")
```

One way around this is to remove the extra buffering by passing `text.width = NA`. Users can do this themselves as part of a list...

``` r
plt(Temp ~ Day | Month, data = aq, legend = list("bottom!", text.width = NA))
```

... but my feeling is that we should enable this automatically for horizontal ("bottom"/"top") legends.

Any objections?

grantmcdermott:
Fixes #434.

@zeileis I followed your suggestion by adding a bit of extra space at the end of each legend label, so that they are closer to their own respective keys (rather than being equidistant between keys). Example:

``` r
#> ℹ Loading tinyplot
tinytheme('clean2')

aq = transform(
  airquality,
  Month = factor(month.name[Month], levels = month.name[5:9])
)

plt(Temp ~ Day | Month, data = aq, legend = "bottom!")
```

As part of this PR, I also addressed some limitations that we had w.r.t. to multicolumn legend support. These should work much better now and also provide an escape hatch for cases with many discrete labels that would otherwise overflow the plot region. I added an example to the "Tips & tricks" vignette along these lines.

``` r
plt(
  weight ~ Time | Chick,
  data = ChickWeight,
  legend = list(ncol = 3),
  type = "l"
)
```

<sup>Created on 2025-07-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,12 @@ where the formatting is also better._
 
 ## Development
 
+### New features
+
+- Improved horizontal legend spacing, as well as multicolumn legend support. A
+  new example in the "Tips & tricks" vignettes demonstrates the latter.
+  (#446 @grantmcdermott)
+
 ### Internals
 
 - Move `altdoc` from `Suggests` to `Config/Needs/website`.


zeileis:
Looks great, thanks! :tada:

PR #454, opened 2025-08-12
handle boxplot without y-variable in tinyplot.default

### LINKED ISSUE #453: Boxplot from one variable
Thank you for an excellent package. 

I’m struggling with boxplots. When I create a boxplot from a single variable, tinyplot uses sequence numbers on the x-axis. I would expect a boxplot to handle single-variable plots in the same way a histogram does.

```
library(tinyplot)

# Plot histogram
plt(Nile, type = 'histogram')

# Attemp to plot boxplot from one variable
plt(Nile, type = 'boxplot')

# Make dummy variable for x axis
x <- cut(seq_along(Nile), c(0, 28, 100))

# Plot boxplot from two variables - CORRECT!
plt(Nile ~ x, type = 'boxplot')
```

zeileis:
Fixes #453 

Like `boxplot(x)` it would be good if `tinyplot(x, type = "boxplot")` created a single boxplot without a grouping variable (and without corresponding labels on the x-axis).

I think that this is not easy to handle in `type_boxplot()` but requires a few simple workarounds in `tinyplot.default()` where the `is.null(y)` cases are handled. I've added:

```
    } else if (type == "boxplot") {
      y = x
      x = rep.int("", length(y))
      xlab = ""
      xaxt = "a"
```

With this both the default and the formula interface work:

```
tinyplot(Nile, type = "boxplot")
tinyplot(~ Nile, type = "boxplot")
```

tinyplot-boxplot

Grouping and faceting still works:

```
tinyplot(~ bill_len | species, data = penguins, type = "boxplot")
```

tinyplot-boxplot

```
tinyplot(~ bill_len, facet = ~ species, data = penguins, type = "boxplot")
```

tinyplot-boxplot

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,10 @@ where the formatting is also better._
 - Improved horizontal legend spacing, as well as multicolumn legend support. A
   new example in the "Tips & tricks" vignettes demonstrates the latter.
   (#446 @grantmcdermott)
+- Univariate boxplots (without grouping variable) are now handled in
+  `tinyplot.default()`, so that `tinyplot(x, type = "boxplot")` and
+  `tinyplot(~ x, type = "boxplot")` essentially produce the same output as
+  `boxplot(x)`. (#454 @zeileis)
 
 ### Internals
 
--- a/R/tinyplot.R
+++ b/R/tinyplot.R
@@ -762,6 +762,11 @@ tinyplot.default = function(
       if (is.null(ylab)) ylab = "Density"
     } else if (type == "function") {
       if (is.null(ylab)) ylab = "Frequency"
+    } else if (type == "boxplot") {
+      y = x
+      x = rep.int("", length(y))
+      xlab = ""
+      xaxt = "a"
     } else if (!(type %in% c("histogram", "barplot"))) {
       y = x
       x = seq_along(x)


grantmcdermott:
Great, thanks for simple workaround @zeileis.

> P.S.: It would be good to modularize this variable handling in the `type_*` functions somehow.

Agree. I worry it will require a larger refactoring b/c these exceptions come before the type-based  `datapoints` construction. OTOH the current logic might just be inertia from the original design. In which case, moving to a modularized approach would kill two birds with one stone by nicely simplifying things.

> P.P.S.: While looking at `data_boxplot()` I noticed that `ymin` and `ymax` are set up but apparently never used. I think this could just be dropped?

I might be mis-remembering (we should test), but I _think_ we need those for the flip logic [here](https://github.com/grantmcdermott/tinyplot/blob/f7774fb752c630908f9398186ce352581c32205d/R/tinyplot.R#L839-877). I have a vague memory of that anyway...

grantmcdermott:
@zeileis Assuming you agree with my comments (which basically amount to "let's tackle these other issues separately"), do you mind adding a NEWS item? We can merge directly afterwards.

zeileis:
OK, good, `NEWS` item added now.

Issue #384, opened 2025-05-04
Simplify `tinyplot_last_call`?


grantmcdermott:
@vincentarelbundock, one for you...

I believe you wrote the following code chunk and TODO marker: https://github.com/grantmcdermott/tinyplot/blob/31b1b055b41033db5a339ce46f80e8afd867a8f9/R/tinyplot.R#L599-L610

We refactored / added `type_density()` back in #284. Does that mean we can action the TODO simplification?

grantmcdermott:
Found the original [discussion comment](https://github.com/grantmcdermott/tinyplot/pull/246#discussion_r1835351093):

> Unfortunately, I was unable to get this to work. I kept getting an error in the `tinyplot.density` tests. The message is obscure, having to do with a `do.call` and a `substitute`. I just can figure this out.
>
> My guess is that this will disappear by itself when we refactor the density code.
>
> In the meantime, I left TODOs in both files with the appropriate save code that we can eventually.

grantmcdermott:
Thanks mate. I'll take a second when I get a chance. Will close then.

PR #460, opened 2025-08-27
evaluate call from tinyplot_add() in parent.frame()

### LINKED ISSUE #459: tinyplot_add does not work inside functions
Currently, `tinyplot_add()` does not work inside function definitions:

```
## example function
lollipop <- function(x, y) {
  tinyplot(x, y, type = "h")
  tinyplot_add(type = "p", pch = 19)
  tinyplot_add(type = "hline")
}
lollipop(1:5, sin(1:5))
## Error in eval(cal) : object 'x' not found
```

The problem is that inside `tinyplot_add()` the correct call `cal` is constructed and then evaluated there: `eval(cal)`. Thus, this looks for `x` and `y` inside `tinyplot_add()` (where they don't exist) and in the enclosing `baseenv()` (where they also don't exist.

In other words, `tinyplot_add()` currently only works correctly when used in the base environment (or when a copy of all arguments are in the base environment).

zeileis:
Fixes #459 

`tinyplot_add()` now evaluates the additional call in the environment from which `tinyplot_add()` is called so that it also works in non-base environments such as in function definitions.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,12 @@ where the formatting is also better._
   `tinyplot(~ x, type = "boxplot")` essentially produce the same output as
   `boxplot(x)`. (#454 @zeileis)
 
+### Bug fixes
+
+- `tinyplot_add()` now evaluates the additional call in the environment from
+  which `tinyplot_add()` is called so that it also works in non-base environments
+  such as in function definitions. (#460 @zeileis)
+
 ### Internals
 
 - Move `altdoc` from `Suggests` to `Config/Needs/website`.
--- a/R/tinyplot_add.R
+++ b/R/tinyplot_add.R
@@ -63,7 +63,7 @@ tinyplot_add = function(...) {
   }
 
   cal[["add"]] = TRUE
-  eval(cal)
+  eval(cal, envir = parent.frame())
 }
 
 


zeileis:
OK, the reason for the failure of some tests is that calls `plt()` store the calls to `tinyplot.formula()` or `tinyplot.default()`:

```
plt(1:5, sin(1:5))
getOption("tinyplot_last_call")
## tinyplot.default(1:5, sin(1:5))
```

Previously, these were found because inside `tinyplot_add()` the namespace of `tinyplot` (containing all methods etc.) is available but to the user only the exported functions are available. So to fix this we need to do one of:

1. Also store calls to `plt()`, not only `tinyplot()`.
2. Replace calls to `tinyplot.*()` methods with `tinyplot()`.
3. Make the entire namespace available again.

I think 1 is the cleanest solution. I'll commit that to see whether it fixes the problems.

zeileis:
This seems to work but made me realize another problem. When the custom function name starts with `tinyplot` things do not work either:

```
tinyplot_lollipop <- function(x, y) {
  tinyplot(x, y, type = "h")
  tinyplot_add(type = "p", pch = 19)
  tinyplot_add(type = "hline")
}
tinyplot_lollipop(1:5, sin(1:5))
## Error in tinyplot_lollipop(1:5, sin(1:5), type = "p", pch = 19, add = TRUE) : 
##   unused arguments (type = "p", pch = 19, add = TRUE)
```

The reason is that the `tinyplot_last_call` after running the first line inside `tinyplot_lollipop()` does not contain

```
tinyplot(x, y, type = "h")
```

but instead

```
tinyplot_lollipop(1:5, sin(1:5))
```

which then gets expanded with `type = "p", pch = 19, add = TRUE` and evaluated which does not work.

So rather than grepping for `^tinyplot` (as in main) or for `(^tinyplot)|(^plt$)` (as in this PR), we should maybe grep for `(^tinyplot$)|(^tinyplot::tinyplot$)|(^plt$)|(^tinyplot::plt)`?

zeileis:
Grant @grantmcdermott & Vincent @vincentarelbundock, I just noticed that my PR is closely related to #384 and #385 which are both still open.

However, after reading both the issue and the PR I'm not 100% sure what the issue is exactly. Is it just to use an environment rather than the global options for storing the last call? If so, I'm happy to give it a shot.

Or if there is more, then let me know and I'll try my best as well :sweat_smile:

grantmcdermott:
> However, after reading both the issue and the PR I'm not 100% sure what the issue is exactly. Is it just to use an environment rather than the global options for storing the last call? If so, I'm happy to give it a shot.

I think that's the correct. I noticed Vincent's FIXME comments and was just circling back. I probably should have closed #385, since it was just a proof of concept where Vincent was demo'ing the fact that the proposed fix didn't work.

That being said, I _do_ think it would be nice to store the previous call in a (the?) dedicated `tinyplot` environment, rather than as a global option... if only for neatness. If you have a path forward here, please give it a go!

P.S. I realise that I'm the bottleneck for quite a few pending changes to `tinyplot` ATM. [dbreg](https://github.com/grantmcdermott/dbreg) and some other projects have been taking up most of my FOSS dev time. Hoping to circle back soon!

grantmcdermott:
Perfect, thanks @zeileis!

zeileis:
:partying_face:

Issue #406, opened 2025-05-19
Dodge `linerange` and friends


vincentarelbundock:
Do we have a good way to "dodge" plots line `linerange` and `errorbar`?

![Image](https://github.com/user-attachments/assets/f011e9d9-e651-46ce-81a7-879bce7d2b45)

grantmcdermott:
Not yet. Implementing this internally wouldn't be too hard, I'm guessing. Are you thinking it would run through the type function itself, e.g. `type_errorbar(dodge = TRUE)`?

vincentarelbundock:
Yes, I think this would be most natural. Perhaps `dodge` would accept a numeric value to control the horizontal space between elements.

strengejacke:
Upvoting! 😬  desperately needed for _modelbased_.

grantmcdermott:
@strengejacke Now available in the dev version thanks to @vincentarelbundock's PR in #461.

PR #461, opened 2025-09-07
`type_pointrange()` and `type_errorange()` dodge

vincentarelbundock:
Issue \#406 `type_linerange()` and `type_errorbar()` get a `dodge` argument.

Default is 0 (no dodge) and can be set to a positive number to dodge overlapping lineranges or errorbars. The dodging amount is a proportion of the width between notches on the x axis (since x is treated as sequential factor/integer positions).

``` r
library(tinyplot)

models <- list(
  "Model A" = lm(mpg ~ wt + cyl, data = mtcars),
  "Model B" = lm(mpg ~ wt + hp + cyl, data = mtcars),
  "Model C" = lm(mpg ~ wt, data = mtcars)
)

results = lapply(names(models), function(m) {
  data.frame(
    model = m,
    term = names(coef(models[[m]])),
    estimate = coef(models[[m]]),
    setNames(data.frame(confint(models[[m]])), c("conf.low", "conf.high"))
  )
})
results = do.call(rbind, results)

tinyplot(estimate ~ term | model,
  ymin = conf.low, ymax = conf.high,
  flip = TRUE, data = results,
  type = type_pointrange(dodge = 0.1))
```

``` r

tinyplot(estimate ~ term | model,
  ymin = conf.low, ymax = conf.high,
  flip = TRUE, data = results[results$model %in% c("Model A", "Model B"), ],
  type = type_errorbar(dodge = 0.4))
```

### DIFF
--- a/R/type_errorbar.R
+++ b/R/type_errorbar.R
@@ -2,10 +2,16 @@
 #'
 #' @description Type function(s) for producing error bar and pointrange plots.
 #'
-#' @param dodge Numeric value (>= 0) for dodging of overlapping points. Dodging
-#'   is scaled relative to the x-axis tick locations (i.e., unique levels of
-#'   `x`). For example, `dodge = 0.5` would place the outermost dodged points
-#'   exactly midway the between axis ticks. Default value is 0 (no dodging).
+#' @param dodge Numeric value (>= 0) for dodging of overlapping `by` groups.
+#'   Dodging is scaled relative to the x-axis tick locations (i.e., unique
+#'   levels of `x`). For example, `dodge = 0.5` would place the outermost dodged
+#'   groups exactly midway the between axis ticks. Default value is 0 (no
+#'   dodging).
+#' @param fixed.pos Logical indicating whether dodged groups should retain a
+#'   fixed relative position based on their group value. Relevant for `x`
+#'   categories that only have a subset of the total number of groups. Defaults
+#'   to `FALSE`, in which case dodging is based on the number of unique groups
+#'   present in that `x` category alone. See Examples.
 #' @inheritParams graphics::arrows
 #' @examples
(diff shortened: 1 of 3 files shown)


grantmcdermott:
Love it. Two quick thoughts: 

1) Should dodging be turned on by default/automatically for grouped pointrange or errorbar plots? I can't think of a situation where I'd like them to overlap (although you probably want to switch off the automatic dodging if `facet_by` is TRUE... Just make sure we pass it through dots from [here](https://github.com/grantmcdermott/tinyplot/blob/ef7ab03ce32cc3f7a02685de6af0dd69d49eaefb/R/tinyplot.R#L856))
2) Can you add a test?

vincentarelbundock:
>     1. Should dodging be turned on by default/automatically for grouped pointrange or errorbar plots? I can't think of a situation where I'd like them to overlap (although you probably want to switch off the automatic dodging if `facet_by` is TRUE... Just make sure we pass it through dots from [here](https://github.com/grantmcdermott/tinyplot/blob/ef7ab03ce32cc3f7a02685de6af0dd69d49eaefb/R/tinyplot.R#L856))

I think default should remain 0. Feels more transparent, and I think that's what `ggplot2` does, so noone will be surprised. Also, I've had groups with non-overlapping intervals several times that I just displayed on the same line, so that's a clear use-case.

>     2. Can you add a test?

Will do.

zeileis:
My intuition would have been the same as yours, Grant. So an option for it would be nice - even if it's not turned on by default.

vincentarelbundock:
I worry that it will look weird when width is large, but I don't have a very strong opinion about this.

Feel free to modify the PR and merge when you're happy with it.

grantmcdermott:
Cool. I'm out ATM but will have a think and then try to implement when I get home.

zeileis:
:partying_face:

Issue #423, opened 2025-06-06
Invalid custom axis title position for `type='barplot'`


lbelzile:
The y-axis label displays on the x-axis instead. This does not depend on the `beside` argument.

``` r
set.seed(2025)
n <- 100L
x <- factor(sample(0:1, size = n, replace = TRUE))
y <- rpois(n, 5)
tinyplot::plt(~ y | x, type = "barplot", beside = TRUE, ylab = "mycaption")
```

![](https://i.imgur.com/0MJ3jdx.png)

<sup>Created on 2025-06-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

lbelzile:
This is from lines 80--82 in file `type_barplot`

where the code states
```
   if (is.null(datapoints$y)) {
          xlab = ylab
          ylab = "Count"
```

zeileis:
Good catch. I think this is just intended to be the default if the user does not specify any `xlab` and/or `ylab`. In that default case it is reasonable.

However, when the user wants to change `xlab`, they have to change `ylab` (as demonstrated by you). And they cannot change `ylab` at all.

At first, I thought that we should use

```
        if (is.null(datapoints$y)) {
          if (is.null(xlab)) xlab = ylab
          if (is.null(ylab)) ylab = "Count"
```

However, then the default does not work anymore because the default `xlab` and `ylab` are inserted by `tinyplot.default` before (https://github.com/grantmcdermott/tinyplot/blob/main/R/tinyplot.R#L755-L774) and hence they are not `NULL` here.

Grant @grantmcdermott, do you know what the "canonical" fix would be here?

grantmcdermott:
Thanks for the report @lbelzile. I took a quick look at this and it turns out to be a trickier solve than I first guessed. I'm slammed at work ATM, so it will have to wait a bit until I have time. But we'll get to it.

PR #462, opened 2025-09-08
Centralized and simplify `xlab and `ylab` processing

vincentarelbundock:
The `xlab` and `ylab` arguments get processed at several different places in the (already very long) `tinyplot()` function. This makes it very difficult to follow, audit, or improve the logic.

This is an attempt to pull out all the processing inside a single helper function called `sanitize_xylab()`. This allows a few improvements for consistency:

1. `tinyplot_add()` the deparsed label from the last plot is printed. test-misc.R
2. `tinyplot(0:10)` prints deparsed "0:10" on both axes instead of "Index". test-type-other.R
3. more consistent use of deparsed names (ex: ribbon)


vincentarelbundock:
@grantmcdermott this is an internal refactoring for maintainability.

The only change to snapshots you'll notice is that when we supply only `x` but not `y` and `tinyplot` sets `y=x` automatically, then we print the label on both axes instead of one label and one `Index`. That seems more descriptive, and it's a corner case anyway, so this is an inconsequential change from the users' perspective.

This is a self-contained PR. I plan to take a more serious look at `tinyplot.R` when I have time, because it seems a bit out of control in terms of length and complexity. But I'll try to do it in small chunks for easier review.

grantmcdermott:
But I agree that the current, nested if-else deparsing logic is convoluted and too finnicky. This is also something that @zeileis has brought up a few times, which has caused headaches downstream (iirc #423 is one such case). So I'm all for improving and simplifying where possible.

vincentarelbundock:
>  is often the first thing I show audiences when comparing `tinyplot` to regular base `plot`

Somehow I had not realized this was base behavior. Makes sense to stick with it then.

All the old tests now pass and I added no new snapshots.

R/tinyplot.R
```diff
@@ -695,28 +695,8 @@ tinyplot.default = function(
...
-  axis_types = c("standard", "none", "labels", "ticks", "axis")
-  axes = match.arg(axes, axis_types)
-  xaxt = match.arg(xaxt, axis_types)
-  yaxt = match.arg(yaxt, axis_types)
-  xaxt = substr(match.arg(xaxt, axis_types), 1L, 1L)
-  yaxt = substr(match.arg(yaxt, axis_types), 1L, 1L)
-  axes = any(c(xaxt, yaxt) != "n")
-  if (is.null(frame.plot) || !is.logical(frame.plot)) frame.plot = all(c(xaxt, yaxt) %in% c("s", "a"))
+  tmp = sanitize_axes(axes, xaxt, yaxt, frame.plot)
+  list2env(tmp, environment())
```
grantmcdermott, line 699: Should we just do `list2env(sanitize_axes(...), environment())` directly to avoid writing the temporary list?

R/tinyplot.R
```diff
@@ -782,40 +759,31 @@ tinyplot.default = function(
   facet_attr = attributes(facet) ## TODO: better solution for restoring facet attributes?
   null_facet = is.null(facet)
 
+  tmp = sanitize_xylab(
+    x = x, xlab = xlab, x_dep = x_dep, xmin_dep = xmin_dep, xmax_dep = xmax_dep,
+    y = y, ylab = ylab, y_dep = y_dep, ymin_dep = ymin_dep, ymax_dep = ymax_dep,
+    type = type)
+  list2env(tmp, environment())
```
grantmcdermott, line 766: Same comment as above; I think it's safer to do this directly than creating a temporary variable in case something goes wrong and the value of `tmp` doesn't get (re)assigned properly.

grantmcdermott (changes requested):
Looks good, thanks. Just one or two minor comments.

vincentarelbundock:
made requested changes and merged. Thanks for the review!

PR #463, opened 2025-09-08
Refactor `draw_legend()`

vincentarelbundock:
The purpose of this PR is to simplify the logic of `draw_legends()` somewhat. The main strategy is to have a cleaner separate of concerns, with a cleaner difference between functions that are called to compute coordinates/settings and functions that are called to create side-effects.

I want to do this in a few more ways and think it will open up big long term benefits, perhaps in the same vein as the `type_*()` refactor.

### DIFF
--- a/R/draw_legend.R
+++ b/R/draw_legend.R
@@ -130,6 +130,7 @@ draw_legend = function(
     #
     ## legend args ----
 
+    outer_side = outer_end = outer_right = outer_bottom = FALSE
     list2env(
       compute_legend_args(
         legend = legend,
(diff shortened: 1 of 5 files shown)


vincentarelbundock:
> I'm a little nervous to split the legend functionality into a serious of nested functions, given how hard tricky it can be to debug and troubleshoot already. 

Oh yeah, totally. 

Honestly, I don't think I would have gotten as good of an appreciation for your point before trying to make that PR work. The execution order with `plot.new()` and friends is SUPER tricky, especially with hooks, etc.

My hope/expectation is that changes like this actually make things easier to navigate, because it'll be easier for us to see when/where there are side effects.

> Minor nits remaining.

Should be done.

grantmcdermott:
Out atm but go for it if you're happy and tests are passing.

PR #464, opened 2025-09-09
`tinyplot.R` cosmetic refactor

vincentarelbundock:
This PR does four things:

1. *Consolidate*: related code was spread all over the function. This PR consolidates related calls in coherent/related chunks.
2. *Comment*: add block comments to clearly highlight groups of related calls
3. *Modularize*: spin off title drawing into `draw_title()` helper function.
4. *Explicit*: make some of the `list2env()` calls more explicit. They are convenient and I was the one who introduced them, but I now feel that they are quite a bad idea, because they obscure which variables are being passed to (and potentially) overwritten by a function. This is a first step toward a future refactor to make internal argument passing easier and more robust. 

I consider this a self-contained contribution ready for review.

### DIFF
--- a/R/tinyplot.R
+++ b/R/tinyplot.R
@@ -637,9 +637,10 @@ tinyplot.default = function(
     ...) {
 
 
-  ###########################
-  # save parameters and calls 
-  ###########################
+  #
+  ## save parameters and calls -----
+  #
+
   par_first = get_saved_par("first")
   if (is.null(par_first)) set_saved_par("first", par())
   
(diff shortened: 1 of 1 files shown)


vincentarelbundock:
FYI, I changed very little code in this PR, but I moved many lines around, which makes review difficult. 

Sorry!

Here's the link to the whole file: https://github.com/grantmcdermott/tinyplot/blob/ef588b88eb3daebe730a284fa5e402437d8a3ffa/R/tinyplot.R

vincentarelbundock:
Thanks for the review @grantmcdermott 

I agree with all your comments and have implemented all of them except for 1: avoid temporary objects and stricter `list2env`.

The reason is that this PR is just the last cosmetic step toward implementing my Master Plan (tm). I'll open a new WIP / spec PR soon with a concrete proposal, and if you like that idea, it'll make all these messy calls and temp variables obsolete.

If it doesn't work out, I can circle back to this and cleanup the residual `tmp` objects.

So I'd just like to merge the cosmetics for now, and move on the actual substantive proposal.

grantmcdermott:
Sgtm. I'm excited to see the master plan!

Feel free to merge if you're happy.

PR #473, opened 2025-09-10
Internal settings container

vincentarelbundock:
This is a more fleshed out example of the centralized internal settings strategy.

Again, no commitment, but I thought it would be useful to look at what it could actually look like. This currently works with just a named list (no S4), but there are 3-4 of the types have not been converted over yet.

@grantmcdermott @zeileis the file of interesting would be this one:

https://github.com/vincentarelbundock/tinyplot/blob/a913e12e45e593110293a663d201bdb247f94179/R/tinyplot.R

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,10 @@ where the formatting is also better._
 
 ## Dev version
 
+### Breaking change
+
+- Internal settings and parameters are now stored in an environment called `settings`, which can be accessed and modified by type-specific functions. This may require changes to users' custom type functions that previously accessed settings as passed arguments. This change was necessary to improve the modularity and maintainability of the codebase, and to add flexibility. (#473 @vincentarelbundock and @grantmcdermott)
+
 ### New features
 
 - `type_text()` gains a `family` argument for controlling the font family,


vincentarelbundock:
CI failures are just warnings. Trivial stuff that would obviously need to be fixed, but the PR still works for illustration.

zeileis:
Thanks, Vincent, for suggesting this. I like the approach with the simple `settings` list and the `sanitize_*()` functions which do `list2env()` and eventually `modifyList()`. This could work...but I didn't work through the details, yet. So far, I only have two more specific (and rather minor) comments:

The reason for including the `raw_input` settings in the settings themselves (https://github.com/vincentarelbundock/tinyplot/blob/internal_settings_v3/R/tinyplot.R#L735) is in order to have easy access to it in the `sanitize_*()` functions without requiring another argument? This feels a bit weird to me (and I don't like the name `raw_input`) but it's a very lean solution and I don't see a real argument against it.

The `sanitize_ribbon.alpha()` function is now a bit different from the other `sanitize_*()` functions because it does not take the whole settings as its argument. Also, it's called within setting up the original input settings (https://github.com/vincentarelbundock/tinyplot/blob/internal_settings_v3/R/tinyplot.R#L730) which differs from the other sanitizers.

vincentarelbundock:
@zeileis two good points. agreed on both counts.

Ideally, I'd like to get rid of `raw_input` altogether, but I'll see if that's realistic.

And yes, a more consistent solution for ribbon alpha should be on my todo list.

vincentarelbundock:
@grantmcdermott and @zeileis

This PR now passes all the tests from `main`, and `R CMD check` passes without notes or warning. I feel we are at a good spot to review and potentially merge.

What is done:

1. `settings` is a named list of internal objects, passed from function to function, and then eventually `list2env()`.
2. Similar operations have been re-organized and consolidated: limits together, axes together, etc.
3. Similar operations are outsourced to helper functions like `sanitize_axes()`, `sanitize_facets()`, `sanitize_datapoints()`, etc.
4. Everything is refactored until (but excluding) `by_col()`, `by_bg()`, facet, and legend handling.

I would like to get this merged before continuing work. Roadmap for the future would include:

1. Refactor and clean-up each of the `sanitize_*()` functions individually. 
    - For now, I have not changed any of the code in there; just moved it around. A lot of this code could be improved and simplified. The beauty of this PR is that it isolates related functionality, so we can work on purely separate PRs to change individual components more easily.
2. Use `settings` for `by_col()` and `by_bg()`.
    - This is tricky because `palette` handling is very fragile at the moment. This needs a re-think.
3. Use `settings` in legends.
    - Be *extra careful*!
    - This is medium/long term because it will difficult, but I think it will make the code more robust and readable.

grantmcdermott:
Thanks @vincentarelbundock. I'm still snowed with other projects, but hoping that I'll be able to review next weekend.

R/by_aesthetics.R
```diff
@@ -1,4 +1,72 @@
...
+    pch = pch,
+    lty = lty,
+    lwd = lwd,
+    cex = cex,
+    col = col,
+    bg = bg
+  )
+}
+
+
```
grantmcdermott: Nit: Can we add a comment (section header) here demarcating the subsidiary functions from the container settings?

R/flip.R
```diff
@@ -0,0 +1,50 @@
...
+      settings = swap_elements(settings, "x", "y")
+      settings = swap_elements(settings, "xaxb", "yaxb")
+      settings = swap_elements(settings, "xaxl", "yaxl")
+      settings = swap_elements(settings, "xaxs", "yaxs")
+      settings = swap_elements(settings, "xaxt", "yaxt")
+      settings = swap_elements(settings, "xlab", "ylab")
+      settings = swap_elements(settings, "xlabs", "ylabs")
+      settings = swap_elements(settings, "xlim", "ylim")
+      settings = swap_elements(settings, "xmax", "ymax")
+      settings = swap_elements(settings, "xmin", "ymin")
```
grantmcdermott: This function should be vectorised, no? Assuming that I'm correct, we could cut down on code (and speed things up a bit) with:

```r
settings = swap_elements(
    settings,
    c("x", "xaxb", ...),
    c("y", "yaxb", ...)
)
```

R/sanitize_axes.R
```diff
@@ -0,0 +1,30 @@
...
+  xaxt = match.arg(xaxt, axis_types)
+  yaxt = match.arg(yaxt, axis_types)
+  xaxt = substr(match.arg(xaxt, axis_types), 1L, 1L)
+  yaxt = substr(match.arg(yaxt, axis_types), 1L, 1L)
+  axes = any(c(xaxt, yaxt) != "n")
+  if (is.null(frame.plot) || !is.logical(frame.plot)) frame.plot = all(c(xaxt, yaxt) %in% c("s", "a"))
+
+
+  new = list(axes = axes, xaxt = xaxt, yaxt = yaxt, frame.plot = frame.plot)
+  settings = modifyList(settings, new, keep.null = TRUE)
```
grantmcdermott: I might be misunderstanding something, but why not use `update_settings()` here?

R/sanitize_datapoints.R
```diff
@@ -0,0 +1,48 @@
...
+  datapoints = data.frame(datapoints)
+  if (nrow(datapoints) > 0) {
+    datapoints[["rowid"]] = seq_len(nrow(datapoints))
+    datapoints[["facet"]] = if (!is.null(facet)) facet else ""
+    datapoints[["by"]] = if (!null_by) by else ""
+  }
+
+  # potentially modified variables
+  new = list(x = x, y = y, xaxt = xaxt, datapoints = datapoints)
+  settings = modifyList(settings, new, keep.null = TRUE)
```
grantmcdermott: Ditto: why not use `update_settings()` here?

R/type_boxplot.R
```diff
@@ -94,7 +94,8 @@ draw_boxplot = function(range, width, varwidth, notch, outline, boxwex, staplewe
 
 
 data_boxplot = function() {
-    fun = function(datapoints, bg, col, palette, null_by, null_facet, ...) {
+    fun = function(settings, ...) {
+        list2env(settings, envir = environment())
```
grantmcdermott: A bit worried about inconsistency here compared to, say, `type_abline` where we explicitly extract only the necessary params from the `settings` container, e.g.

```r
list2env(settings[c("datapoints", "bg", "col", "palette", "null_by", "null_facet")], environment())
```

We should decide on one approach and do that for all types.

R/type_boxplot.R
```diff
@@ -138,7 +141,7 @@ data_boxplot = function() {
         if (length(unique(datapoints$by)) > 1) out[["by"]] = datapoints$by
         if (length(unique(datapoints$facet)) > 1) out[["facet"]] = datapoints$facet
 
-        return(out)
+        do.call(update_settings, c(list(settings), out))
```
grantmcdermott: Could we not just call

```
update_settings(settings, out)
```

?

R/type_hline.R
```diff
@@ -4,7 +4,9 @@
 #' @export
 type_hline = function(h = 0) {
   assert_numeric(h)
-  data_hline = function(datapoints, lwd, lty, col, ...) {
+  data_hline = function(settings, ...) {
+    list2env(settings, environment())
```
grantmcdermott: Same

R/type_rug.R
```diff
@@ -33,12 +33,15 @@
...
-  data_rug = function(datapoints, ...) {
+  data_rug = function(settings, ...) {
+    list2env(settings["datapoints"], envir = environment())
     if (nrow(datapoints) == 0) {
       msg = "`type_rug() only works on existing plots with x and y data points."
       stop(msg, call. = FALSE)
     }
-    return(datapoints)
+
+    out = modifyList(settings, list(datapoints = datapoints), keep.null = TRUE)
```
grantmcdermott: Not `update_settings(...)`?

grantmcdermott (commented):
Thanks for this @vincentarelbundock... I finallly had time to review.

I've let a quite a few comments. Some of them are nits, some are clarifications, and some are a suggestions for potential improvements.

I have two high-level observations:

1. This container approach is cool, but I think we can tighten up on consistency and (hopefully) efficiency. One concern I have is the fact that we're basically lugging this `settings` item throughout the internal code, but "manually" reassigning certain elements each time, e.g. `settings = update_settings(settings, list(x = x))`. In that sense, it's really a glorified wrapper around `modifyList` (which is fine!). But perhaps we could use the fact that the target is always the same---i.e, `settings`---to tighten up the implementation so that the code becomes `update_settings(x = x)`? I don't know whether we should enable update by reference, but it could simplify the code further. 

2. Following on from pt. 1, I'm a little uncomfortable with the fact that many of our internal functions now collapse down to taking a single `settings` argument, without explicitly making clear _which_ settings are going to be extracted or adjusted. (I mean, you can obviously look inside the relevant function, but everything is more hidden.) I don't have an obviously better solution, but perhaps we can tweak the implementation so that instead of, say,

```
data_abline = function(settings, ...) {
    list2env(settings[c("datapoints", "lwd", "lty", "col")], environment())
...
```

we end up with

```r
data_abline = function(params = c("datapoints", "lwd", "lty", "col"), ...) {
    list2env(settings[params], environment())
...
```

Or, if you wanted something fancier, you could extract the function args as strings:

```r
data_abline = function(datapoints, lwd, lty, col), ...) {
    call = match.call()
    param_names = setdiff(names(formals()), "...")
    params = sapply(call[param_names], deparse1)
    list2env(settings[params], environment())
...
```

R/flip.R
```diff
@@ -0,0 +1,50 @@
+swap_elements = function(lst, a, b) {
+  if (any(!c(a, b) %in% names(lst))) {
+    out = lst
+  } else if (all(c(a, b) %in% names(lst))) {
+    out = do.call(update_settings, c(list(lst), stats::setNames(lst[c(b, a)], c(a, b))))
```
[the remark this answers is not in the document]
vincentarelbundock: 612133f1f5a50b407af87e4b675c2904acbb31b2

R/type_barplot.R
```diff
@@ -78,9 +78,9 @@ type_barplot = function(width = 5/6, beside = FALSE, center = FALSE, FUN = NULL,
 
 #' @importFrom stats aggregate
 data_barplot = function(width = 5/6, beside = FALSE, center = FALSE, FUN = NULL, xlevels = NULL, xaxlabels = NULL, drop.zeros = FALSE) {
-    fun = function(datapoints, col, bg, lty, lwd, palette, xlab = NULL, ylab = NULL, xlim = NULL, ylim = NULL, xaxt = NULL, yaxl = NULL, yaxt = NULL, axes = TRUE, null_by, facet_by, ...) {
+    fun = function(settings, ...) {
+        list2env(settings[c("datapoints", "xlab", "ylab", "null_by", "facet_by", "xlim", "ylim", "raw_input", "col", "bg", "yaxl", "xaxt")], environment())
```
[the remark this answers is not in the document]
vincentarelbundock: the `data_barplot()` can directly override that specific argument when it calls `update_settings()`

R/by_aesthetics.R
```diff
@@ -427,32 +495,22 @@ by_cex = function(ngrps, type, bubble = FALSE, cex = NULL) {
...
-    by,
-    by_continuous,
-    by_ordered,
-    col,
-    fill,
-    ngrps,
-    palette,
-    ribbon.alpha,
-    type) {
+by_bg = function(bg, fill, col, palette, alpha, by_ordered, by_continuous, ngrps, type, by, ribbon.alpha, adjustcolor) {
```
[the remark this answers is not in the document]
vincentarelbundock: 1fdef88979ce9e88a07ab037adb53811526d5484

R/sanitize_axes.R
vincentarelbundock: b0a1e07f43df0a55f0ae635ffda532e4e0422bbd

R/sanitize.R
```diff
@@ -3,137 +3,3 @@ sanitize_ribbon.alpha = function(ribbon.alpha) {
   if (is.null(ribbon.alpha)) ribbon.alpha = .tpar[["ribbon.alpha"]]
   return(ribbon.alpha)
 }
```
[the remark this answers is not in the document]
vincentarelbundock: 9747199de2cea4fd96bbfd36f8f5d852e509ea8c

R/flip.R
grantmcdermott: Cool cool. I think this new `settings` approach gives us most of that anyway now

PR #484, opened 2025-09-17
Ephemeral `theme` arg

### LINKED ISSUE #442: Support ephemeral themes via `plt(..., theme = ...)`?
Currently, we require that users (un)set themes globally and in persistent fashion via

```r
tinytheme("<theme>")
```

But there are also cases where it would be useful to set ephemeral themes that don't persist. (An example: For a package that calls tinyplot as a dependency, where they would like a particular styling for their plots, but not have to override the user's global settings.)

Perhaps we should enable this via an additional argument in the main function:

```r
tinyplot(..., theme = "<theme>")
```

?

grantmcdermott:
Closes #442

MWE

``` r
pkgload::load_all('~/Documents/Projects/tinyplot')
#> ℹ Loading tinyplot
plt(bill_dep ~ bill_len | species, data = penguins)
```

``` r
plt(bill_dep ~ bill_len | species, data = penguins, theme = 'clean')
plt_add(type = 'lm')
```

``` r
plt(bill_dep ~ bill_len | species, data = penguins)
```

<sup>Created on 2025-09-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,9 @@ where the formatting is also better._
   `boxplot(x)`. (#454 @zeileis)
 - `type_errorbar()` and `type_point_range()` get a `dodge` argument.
   (#461 @vincentarelbundock)
+- The new `tinyplot(..., theme = <theme>)` argument enables users to invoke
+  ephemeral themes as an alternative to the persistent themes that follow
+  `tinytheme(<theme>)`. (#483 @grantmcdermott)
 
 ### Bug fixes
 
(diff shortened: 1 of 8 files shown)


zeileis:
Very nice, thanks for adding this.

I played around a bit how this interacts with `tinyplot_add()` or setting `add = TRUE`. I wasn't able to create weird or unexpected results, though, even when using different themes in the first and in the added plot. :+1:

PR #504, opened 2025-10-06
make tinyplot_add compatible with do.call(tinyplot)

### LINKED ISSUE #499: do.call tinyplot_add
Hey tinyplot maintainers, first, thank you for this excellent package, which makes plotting a delight.

I recognized since version `0.5.0` do.call does not work any more in combination with tinyplot_add (in version `tinyplot_0.4.2` it did work as expected).

Below is a minimal example to reproduce the error. Note you have to start a fresh R session where tinyplot was not used jet.

```r
library("tinyplot")

df <- data.frame(x = 1:10, y = (1:10)^2)

# Raises no previous tinyplot call found error.
do.call(tinyplot, list(y ~ x, data = df, type = "p"))
tinyplot_add(type = "l", data = df)
#R> Error in tinyplot_add(type = "l", data = df) :
#R>   No previous tinyplot call found.

# Workes fine.
tinyplot(y ~ x, data = df, type = "p")
tinyplot_add(type = "l", data = df)
```

The issue seams to be that the following regex does not match when using do.call
```r
    calls = sys.calls()
    tinyplot_calls = "(^tinyplot$)|(^tinyplot::tinyplot$)|(^plt$)|(^tinyplot::plt)|(^tinyplot:::)"
    idx = grep(tinyplot_calls, sapply(calls, function(k) k[[1]]))
```
since sys.calls() gives in this case
```
[[1]]
do.call(tinyplot, list(y ~ x, data = df, type = "p"))

[[2]]
(function (x, ...) 
{
    UseMethod("tinyplot")
})(y ~ x, data = list(x = 1:10, y = c(1, 4, 9, 16, 25, 36, 49, 
64, 81, 100)), type = "p")

[[3]]
tinyplot.formula(y ~ x, data = list(x = 1:10, y = c(1, 4, 9, 
16, 25, 36, 49, 64, 81, 100)), type = "p")

[[4]]
tinyplot.default(x = x, y = y, by = by, facet = facet, facet.args = facet.args, 
    data = data, type = type, xmin = mf[["(xmin)"]], xmax = mf[["(xmax)"]], 
    ymin = mf[["(ymin)"]], ymax = mf[["(ymax)"]], xlim = xlim, 
    ylim = ylim, main = main, sub = sub, xlab = xlab, ylab = ylab, 
    ann = ann, axes = axes, frame.plot = frame.plot, asp = asp, 
    grid = grid, legend_args = legend_args, pch = pch, col = col, 
    lty = lty, lwd = lwd, restore.par = restore.par, ...)
```

You might could use something like that instead of the grep, since comparing objects should
be more reliable.

```
is_tinyplot_method <- function(x) {
  fun <- try(eval(x[[1]]), silent = TRUE)
  identical(tinyplot::tinyplot, fun)
}
idx <- which(as.logical(lapply(calls, is_tinyplot_method)))
```

Hope this helps.

zeileis:
Fixes #499 

The `.last_call` of the `tinyplot()` generic is now determined by comparing the _functions_ called in `sys.calls()` rather than their _names_ as suggested by @FlorianSchwendinger.

This makes it compatible with usage in `do.call()` or when using a custom copy of the generic.

```
do.call(tinyplot, list(x = 1:5, y = sin(1:5)))
do.call(tinyplot_add, list(type = "h"))

myplot = tinyplot
myplot_add = tinyplot_add
myplot(x = 1:5, y = sin(1:5))
myplot_add(type = "h")
```

tinyplot-add

### DIFF
--- a/R/tinyplot.R
+++ b/R/tinyplot.R
@@ -652,10 +652,10 @@ tinyplot.default = function(
   assert_logical(add)
   if (!add) {
     calls = sys.calls()
-    tinyplot_calls = "(^tinyplot$)|(^tinyplot::tinyplot$)|(^plt$)|(^tinyplot::plt)|(^tinyplot:::)"
-    idx = grep(tinyplot_calls, sapply(calls, function(k) k[[1]]))
+    is_tinyplot_call <- function(x) identical(tinyplot, try(eval(x[[1L]]), silent = TRUE))
+    idx = which(vapply(calls, is_tinyplot_call, FALSE))
     if (length(idx) > 0) {
-      set_environment_variable(.last_call = calls[[idx[1]]])
+      set_environment_variable(.last_call = calls[[idx[1L]]])
     }
   }
 


grantmcdermott (changes requested):
Thanks @zeileis, this looks good to me. One minor nit and then we can can merge once #473 is shepherded through. (I want to wait for that, since it's a big change that will cause some adjustments across most of the codebase.)

PR #528, opened 2025-11-25
Expand dodge support to `"p"` and `"l"` types

grantmcdermott:
Mostly for layering, since I know that, say, joined errorbar plots are pretty common. (Even though I'm not the biggest fan, personally.)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@ where the formatting is also better._
 
 ## Dev version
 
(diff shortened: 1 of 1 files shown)


grantmcdermott:
P.S. Small breaking change bundled with this PR: I've renamed the `fixed.pos` argument to `fixed.dodge` to avoid ambiguity.

PR #535, opened 2025-12-14
abstract for Psychoco 2026

zeileis:
In 2026 I would like to present `tinyplot` at a couple of workshops/conferences. The first one is [Psychoco 2026](https://www.psychoco.org/2026/), a moderately-sized psychometric computing workshop that I have been co-organizing for a number of years now. I would list us three (Grant, Vincent, and myself) as authors and myself as the presenter. I've added an abstract in parallel to the useR! 2025 abstract (in a `vignettes` sub-folder). Let me know if this is ok for you or you have any objections.

### DIFF
--- a/inst/Psychoco2026/README.md
+++ b/inst/Psychoco2026/README.md
@@ -1,6 +1,12 @@
-tinyplot: Lightweight Extension of the Base R Graphics System
+## tinyplot: Lightweight Extension of the Base R Graphics System
 
-Grant McDermott, Vincent Arel-Bundock, Achim Zeileis
+_Grant McDermott, Vincent Arel-Bundock, Achim Zeileis_
+
+**Presented at:** Psychoco 2026 - International Workshop on Psychometric Computing, Università di Padova, Italy, February 5-6.
+
+**PDF slides:** <https://www.zeileis.org/papers/Psychoco-2026.pdf>
+
+**Abstract:**
 
 The base R graphics system provides a lot of powerful infrastructure for drawing
 data visualizations. At the core is the `plot()` generic function with its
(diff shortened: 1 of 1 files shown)


vincentarelbundock:
This is fantastic!

grantmcdermott:
SGTM!

I world possibly tweak the abstract a bit (happy to make a few suggestions, but that's only if you feel like indulging my fussiness).

Regarding location, if we're going to add a few presentations then I would think of creating a dedicated "Presentations" slot on the website. Or, treat useR! as special and keep that one in vignettes and add the rest to inst/ with . Rbuildignore instructions?

zeileis:
Sure, feel free to make tweaks, just push them into the repository. I can still rephrase if needed.

Re: dedicated directory. For now I would probably leave things as they are and wait a bit more. But if you want to do something more elaborate, that's very welcome, too.

grantmcdermott:
Slides look great!

I wouldn't worry about adhering to my theme; I think your original PDF is what we want (and may also simplify the _`extensions` part).

I think it makes sense for the useR! presentation to remain as a vignette (and available on the website). For these other presentations, should we just add them to `inst/` and `.Rbuildignore`?

zeileis:
For me either solution is fine: 

- Put presentations in `vignettes/` and including them on the web page.
- Put presentations in `inst/` + `.Rbuildignore` and only have useR! 2025 on the web page.

In the former case I'm still not sure whether I should use the quarto+beamer (rather than quarto+revealjs) code. I think it's easier to re-use for you (and potentially others) without our university's beamertheme.

zeileis:
OK, good, done. I've kept the revealjs format and included the `_extensions`.

How would you tweak slides 2 and 9 to fit the page?

grantmcdermott:
Ah, for you. Honestly, I'm not worried about the university theme. I think it's totally fine to keep the original slides. I think pushing the basic scaffolding (e.g. just the .qmd file and .pdf) is also fine. We don't need to re-render everything "as-is". Just having the source would be enough for me to adapt if needed :-)

PR #536, opened 2025-12-15
Settings legend

vincentarelbundock:
Continued work on removing core computation steps from the `tinyplot.R` file. Here, I created helper functions to draw the legends, and kept only minimal machinery inside `tinyplot()`. 

I also reorganized the code a bit, created a couple helper functions to reduce duplication, and renamed a couple things for consistency and clarity.

I'm sure there are improvements available, but the goal is separation of concern. Then, we can think about simplification/performance/clarity.

Let me know what you think.

PR #541, opened 2026-01-12
Fix interactive resizing regression from Issue #536

### LINKED ISSUE #540: Regression: Legend margins are broken
Follow up to #536 (I think)

I'm not entirely sure why it didn't get flagged by the test suite (probably b/c it's an interactive issue), but my legends are being cut off in Positron now.

Image

Resizing the plot window makes the problem worse.

Image

This looks very much like a regression of https://github.com/grantmcdermott/tinyplot/pull/438 In other words, my guess is that some of the settings aren't being passed correctly to `recordGraphics()`. Or maybe the `recordGraphics()` logic doesn't work with a container environment?

@vincentarelbundock are you seeing similar behaviour? Unfortunately, I won't have time to look at this for a few days. But if we can't fix quickly, then we should revert #536 ASAP.

vincentarelbundock:
Fix #540

Problem                                                               
                                                                        
  PR #438 fixed a long-standing issue where resizing the plot window    
  would cause legends to become misaligned. The fix worked by wrapping  
  coordinate-dependent calculations inside recordGraphics(), which R    
  replays when the device is resized.                                   
                                                                        
  PR #536 refactored the legend code but inadvertently broke this fix.  
  The refactoring moved critical calculations outside of                
  recordGraphics():                                                     
          
```r                                                              
  # These now happen OUTSIDE recordGraphics - only run once at initial  
  size:                                                                 
  legend_outer_margins(legend_env, apply = draw)     # margin adjustment
  legend_env$inset = measure_legend_inset(legend_env) # inset           
  calculation                                                           
  plot.new()  # plot refresh                                            
                                                                        
  # INSIDE recordGraphics - only this replays on resize:                
  recordGraphics({                                                      
    do.call("legend", legend_env$args)  # draws with STALE inset values 
  }, ...)                                                               
                      
```
                                                  
  When the window is resized, only the final legend() call replays, but 
  it uses margin/inset values calculated for the original window size.  
                                                                        
  Fix                                                                   
                                                                        
  Move the margin application, inset calculation, and plot refresh back 
  inside recordGraphics() via a new tinylegend() helper function:       
                                  
```r                                      
  # Initial setup only (first plot.new, measurement):                   
  legend_outer_margins(legend_env, apply = FALSE)                       
                                                                        
  # Store base values for replay:                                       
  legend_env$omar_base = legend_env$omar                                
  legend_env$ooma_base = legend_env$ooma                                
  legend_env$inset_base = legend_env$args[["inset"]]                    
                                                                        
  # Everything else inside recordGraphics:                              
  recordGraphics(                                                       
    tinylegend(legend_env),  # recalculates margins, inset, then draws  
    list = list(legend_env = legend_env),                               
    env = getNamespace("tinyplot")                                      
  )                                                                                                                              
```
               
  The tinylegend() function resets to base values on each replay, then  
  recalculates soma, applies margins, calculates inset, and             
  draws—ensuring correct positioning after resize.

### DIFF
--- a/R/legend.R
+++ b/R/legend.R
@@ -205,6 +205,85 @@ measure_legend_inset = function(legend_env) {
 }
 
 
+# Internal workhorse function for legend rendering
+# This function is called inside recordGraphics() so that all coordinate-dependent
+# calculations are re-executed when the plot window is resized
+tinylegend = function(legend_env) {
+  # Reset to base values (before soma/inset were applied)
+  # This is necessary because legend_env is passed by reference and modifications
+  # persist across recordGraphics() replays
+  legend_env$omar = legend_env$omar_base
+  legend_env$ooma = legend_env$ooma_base
+  legend_env$args[["inset"]] = legend_env$inset_base
+
+  # Re-measure legend dimensions (device size may have changed on resize)
+  legend_env$dims = measure_fake_legend(legend_env)
+
+  # Calculate and apply soma (outer margin adjustment based on legend size)
+  soma = if (legend_env$outer_side) {
+    grconvertX(legend_env$dims$rect$w, to = "lines") - grconvertX(0, to = "lines")
+  } else if (legend_env$outer_end) {
+    grconvertY(legend_env$dims$rect$h, to = "lines") - grconvertY(0, to = "lines")
+  } else {
+    0
+  }
+  soma = soma + sum(legend_env$lmar)
+
+  if (legend_env$outer_side) {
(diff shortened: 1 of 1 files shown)


vincentarelbundock:
Disclaimer: This solution was 1-shotted by Claude Code after I fed it `.diff` files for the two PRs mentioned above.

I have reviewed the code. The changes seem pretty straightforward and, AFAICT, "correct".

I have also tested it locally and resizing the Positron window no longer cuts-off the legend.

PR #550, opened 2026-02-21
Remove duplicated `swap_columns()`

etiennebacher:
Hi, I tested a new linter rule on this package and found that `swap_columns()` was defined once in `flip.R` and once in `utils.R`, with the same implementation. I removed the former (just because I had to pick one).


vincentarelbundock:
Thanks, this is great!

PR #558, opened 2026-03-22
tinyplot method for 'ts' time series objects

zeileis:
In order to visualize `ts` time series with `tinyplot` it would be nice to have a dedicated method. Internally, this would just set up a long data frame, e.g., with variables `Time`, `Value`, and `Series` and then call `tinyplot()` with a suitable formula.

As you know, I previously already created a similar method in our `zoo` package, see: <https://github.com/cran/zoo/blob/master/R/tinyplot.zoo.R>. And for my personal usage it was enough to do `tinyplot(as.zoo(x))` if `x` is a `ts` series. However, currently I'm teaching my time series course where I still work with quite a few `ts` series as well and felt it would be nice to have a native `ts` method.

Hence, this PR suggests such a method. I will need a few more commits to set up tests and news, etc. I will also post some examples in this PR...hopefully later today.


grantmcdermott:
Thanks. @zeileis, this is very cool. I'm still in the middle of busy path, but had some FOSS time tonight so had a peek and can give you my quick thoughts:

1. Agree that we stay with idiomatic `tinyplot` defaults. I'd encourage single column facets via the standard `facet.args(ncol = 1)` argument now... Although I like your idea of `facet = Series ~ 1`. My feeling is that this requires a separate PR, though.
2. Yes, ideally we'd allow `by` too. In fact, again thinking about expected `tinyplot` behaviour, my sense is that we should group `by` default instead of  diverting to `facets`. I realise that this is a break with `plot`, but I'd prefer for us to maintain internal consistency if we had to break with one or the other. (Maybe the default is `tinyplot.ts(..., facet = "by")`?)
3. I haven't tested locally, so I haven't seen the other failure. But I'd very much like to preserve `tinyplot(Nile, type = "histogram")` behaviour. I haven't thought this through carefully, but could we maybe skip the internal `Value ~ Time (| Series)` formula validation if the user provides an explicit (non-NULL) `type`?

grantmcdermott:
Getting back to this (since I missed my v0.7.0 deadilne anyway 🥴 )... Here's where I think we stand:

1. The two `Nile` test snapshots are still failing. Again, I'd really prefer to keep the current (compatible with base R) behaviour for `plt(Nile, type = "histogram")`. I'll try to think on a solution.
2. Point taken about `facets` as a better default than `by` (at least on its own). I guess my primary non-faceted use case is direct labels, but this can be enabled with, say, `plt(EuStockMarkets, facet = NULL, legend = list('direct', repel = TRUE)`. We should just add an example along these lines to the docs.

grantmcdermott:
Took a little more effort than I figured last night, but got there in the end. Thanks again for pushing this 🎉

grantmcdermott:
P.S. Here's the rendered man page on the website: https://grantmcdermott.com/tinyplot/man/tinyplot.ts.html

PR #562, opened 2026-03-24
Support facet formulas y~1 and 1~x for 1-column or 1-row layouts

zeileis:
In #558 we discussed the idea that `facet = y ~ 1` could be used as a shortcut for `facet = ~ y, facet.args = list(nrow = length(unique(y)))`. Analogously, `facet = 1 ~ x` could be used as a shortcut for `facet = ~ x, facet.args = list(nrow = 1)`.

I have implemented this now by tweaking `tinyframe()` slightly. This now distinguishes `~ x` from `1 ~ x` by returning `NULL` for the `yfacet` in the former case but a zero-column data frame in the latter case. In other words, a zero-column data frame signals: This part of the formula had a specification but no variables in it. In `tinyplot.formula()` I've extended the processing of `xfacet` and `yfacet` correspondingly.

In principle, this works but there are two caveats:

1. `tinyplot.default()` handles the case with `facet.args` slightly differently from facet attributes.
2. The processing has only been extended in the formula method. I just noticed now that the is an additional function `get_facet_fml()` (https://github.com/grantmcdermott/tinyplot/blob/facet-formula/R/facet.R#L613-L658) that does something similar to my `tinyformula()`/`tinyframe()` processing but is only used in the default method. We should probably try to use the same code in both cases.

To illustrate the first point, consider

```
library("tinyplot")
tinytheme("clean2")
p = transform(penguins, group = factor(paste(species, island, sep = "-")))
tinyplot(bill_dep ~ bill_len, data = p, facet = 1 ~ group)                            ## top
tinyplot(bill_dep ~ bill_len, data = p, facet = ~ group, facet.args = list(nrow = 1)) ## bottom
```

facet2

```
tinyplot(bill_dep ~ bill_len, data = p, facet = group ~ 1)                            ## left
tinyplot(bill_dep ~ bill_len, data = p, facet = ~ group, facet.args = list(nrow = 5)) ## right
```

facet1

Note that in the formula-based specification we get the right facet strips as well. Probably this should only be done in case both the number of rows and the number of columns is greater than 1? I wasn't sure though where to best add this information.


zeileis:
Sure, perfectly fine! :rocket:

zeileis:
I have addressed the problems above as follows now:

- The route via `facet_grid`/`facet_nrow` attributes is only used for `facet = y ~ x` as before!
- The case of `facet = 1 ~ x` and `y ~ 1` is handled without attributes and using `facet.args = list(nrow = ...)`.

Now these new specifications work as desired:

```
library("tinyplot")
tinytheme("clean2")
p = transform(penguins, group = factor(paste(species, island, sep = "-")))
tinyplot(bill_dep ~ bill_len, data = p, facet = 1 ~ group)
```

tinyplot-facet1


```
tinyplot(bill_dep ~ bill_len, data = p, facet = group ~ 1)
```

tinyplot-facet2

And these old specifications work as before:

```
tinyplot(bill_dep ~ bill_len, data = p, facet = ~ group)
```

tinyplot-facet3

```
tinyplot(bill_dep ~ bill_len, data = p, facet = island ~ species)
```

tinyplot-facet4


Notes:

- If the user specifies something like `facet = 1 ~ x, facet.args = list(nrow = n)`, then it is handled like  `facet = ~ x, facet.args = list(nrow = n)`. Thus, the user-specified `nrow` is used (without warning) and not the `nrow` implied by the `1 ~ x` formula.
- Test cases with snapshots have been added (expanding the ozone examples in `test-facet.R`).

grantmcdermott:
Thanks @zeileis!

zeileis:
Great, thanks, Grant! I hadn't added a NEWS item, yet. Do you want to do so or should I? If so where/how?

Re: `get_facet_fml()` (https://github.com/grantmcdermott/tinyplot/blob/facet-formula/R/facet.R#L613-L658). Should I create an issue for this so that we remember to have a look at replacing it with `tinyformula()`?

grantmcdermott:
Yes to both, thanks Achim. (For NEWS, I'm fine if you push direct to `main` once you've synced the latest changes.)

grantmcdermott:
Had a sec this morning, so I just pushed through the NEWS entry in https://github.com/grantmcdermott/tinyplot/commit/7d37682fd53cfe8f17cea8a744dbcbee515024ab.

But please still raise an issue for the below when you get a chance:

> Re: `get_facet_fml()` (https://github.com/grantmcdermott/tinyplot/blob/facet-formula/R/facet.R#L613-L658). Should I create an issue for this so that we remember to have a look at replacing it with `tinyformula()`?

Issue #555, opened 2026-02-24
`asp` not respected when `facet.args = list(free = TRUE)`


katrinabrock:
### MRE

This creates a 1x1 square with a length 10 line segment protruding to the right. `asp=1` should make the square look like a square, not a rectangle. "A", "B", and "C" have identical data.

```r
png(width = 600, height = 200, unit = "px")
tinyplot(
  y ~ x,
  facet = ~group,
  facet.args = list(free = TRUE),
  data = data.frame(
    y = rep(c(1, 1, 2, 2, 1, 1.5, 1.5), 3),
    x = rep(c(2, 1, 1, 2, 2, 2, 11), 3),
    group = c(rep("A", 7), rep("B", 7), rep("C", 7))
  ),
  type = "l",
  asp = 1
)
dev.off()
```

### Expected behavior

Since the data is identical for each group, `facet.args = list(free = TRUE)` should have no impact. Different plots should be free show different ranges, but `asp=1` should be preserved within each plot.

<img width="600" height="200" alt="Image" src="https://github.com/user-attachments/assets/3d997084-7265-49c3-97ad-49d880e43244" />

### Actual behavior

<img width="600" height="200" alt="Image" src="https://github.com/user-attachments/assets/81a1aa06-055c-4bdc-95cb-363f9bad40ba" />

PR #568, opened 2026-04-02
Issue 555

katrinabrock:
I've made an attempt to solve issue #555 , but it's not quite working. The plot I'm getting looks like this:

image

<details> <summary> diff plot </summary>

image

</details>

So pane A is almost right and pane B is further off.

I've diagnosed that when `usr` is set manually, `asp` is not taken into account. https://github.com/grantmcdermott/tinyplot/blob/ba467917db69f25160a07f65cc6c5d249a628082/R/facet.R#L325-L328

I've attempted to translate [the logic used by plot.window()](https://github.com/wch/r-source/blob/8fec469cb5de80c3e6d7feae4d906180a39c5426/src/library/graphics/src/plot.c#L548-L575). However, it's not working. I believe the issue is that `grconvertX(1, "npc", "inches")` is not actually equivalent to `GConvertXUnits(1.0, NPC, INCHES, dd)` or maybe I'm calling `grconvertX` in the wrong context.

### DIFF
--- a/R/facet.R
+++ b/R/facet.R
@@ -311,8 +311,9 @@ draw_facet_window = function(
         # individual facet.
         xfree = if (!is.null(facet)) split(c(x, xmin, xmax), facet)[[ii]] else c(x, xmin, xmax)
         yfree = if (!is.null(facet)) split(c(y, ymin, ymax), facet)[[ii]] else c(y, ymin, ymax)
-        if (null_xlim) xlim = range(xfree, na.rm = TRUE)
-        if (null_ylim) ylim = range(yfree, na.rm = TRUE)
+        lim = calc_lim(range(xfree, na.rm = TRUE), range(yfree, na.rm = TRUE), asp)
+        if (null_xlim) xlim = lim$xlim
+        if (null_ylim) ylim = lim$ylim
         xext = extendrange(xlim, f = 0.04)
         yext = extendrange(ylim, f = 0.04)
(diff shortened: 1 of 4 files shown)


katrinabrock:
Ok, this seems very strange to me. This is totally outside tinyplot context:

```r
print_scale <- function(from = "npc", to = "inches") print(c(
  grconvertX(1, from, to),
  grconvertY(1, from, to)
))

print_scale()

print("set mfrow")
par(mfrow = c(1,2))
print_scale()

print("mfg=c(1,1)")
par(mfg = c(1,1))
print_scale()

print("mfg=c(1,2)")
par(mfg = c(1,2))
print_scale()
```

Result:
```
[1] 6.58 6.18
[1] "set mfrow"
[1] 6.58 6.18
[1] "mfg=c(1,1)"
[1] 3.08 6.18
[1] "mfg=c(1,2)"
[1] 6.58 6.18
```

It seems very odd that setting the first figure vs the second figure to be drawn next would impact npc->inches conversion.

@grantmcdermott or @zeileis I would love some input here. I read #65 which seems related, but I still feel like there's something fundamental I'm missing here.

grantmcdermott:
Sorry for the slow reply @katrinabrock. #65 does feel related, but honestly I haven't deep dived the issue to confirm. I'm trying to shepherd through some internal theme refactoring logic ATM and will return to these other open PRs once that is merged.

grantmcdermott:
Superseded by #744

Issue #436, opened 2025-07-03
Support for categorical variables in type_summary()


marcboschmatas:
I think it would be interesting to allow for the support of categorical variables in the x axis for summary plots, similar to ggplot, so that you could have something like 

```
plt(bill_length_mm ~ species, data = penguins, type = type_summary(mean_se)
```

And get this in return

![Image](https://github.com/user-attachments/assets/06247029-1a5c-435d-b495-2556608605df)

grantmcdermott:
Thanks for the suggestion. The tricky thing is that you'd need to amend `type_summary` so that it does two new things: (a) understands array returns with y, ymin, and ymax, and (b) can dynamically switch drawing types instead of just reverting to `type_lines()`. 

(a) would require a major refactor since we couldn't just rely on `ave()` anymore for the internal conversion. (b) would be complicated on its own, although potentially doable via a `draw_type` argument or similar. 

Given this, my guess is that it would just be easier to call `plt(..., data = mean_se(penguins), type = "pointrange")`.

Issue #382, opened 2025-05-01
`"coefplot"`


grantmcdermott:

vincentarelbundock:
FWIW, I have done this in `modelsummary` and it is quite tricky to do with generality. I don't think we want to support all possible packages, and a type that only supports `glm` and base fitting functions would be pretty bad and uncompetitive.

vincentarelbundock:
One possibility would be a Suggests on `parameters` package. Not sure you want that.

grantmcdermott:
Thanks @vincentarelbundock; definitely interested in your experiences here (I was hoping you'd see this issue and chime in). FWIW I was busy writing down my own thoughts before your reply, so here they are for posterity:

I've been wondering whether we should support a `"coefplot"` type (i.e., as convenience layer which could then get passed to `"pointrange"`, `"errorbar"`, or `"ribbon"`). We would obviously need to decide on the scope and limitations, including supported model engines (`lm` as default, etc.) But in principle I wonder whether we could support any model that `coef` and `confint` methods?

Alternatively (or in addition), it might be worth PRing `tinyplot` methods for the **broom** and **parameters** repos.

grantmcdermott:
> One possibility would be a Suggests on `parameters` package. Not sure you want that.

Yeah, I wouldn't do that on our side. But we could propose a `parameters_model.tinyplot`method on their side that would kick in if both packages are loaded.

zeileis:
Same for `modelsummary` and `coeftest` objects?

vincentarelbundock:
```r
plt(am ~ wt, mtcars, type = type_coefplot(model = glm, args = list(family = 'binomial'), style = 'errorbar')) # customized ex.
```

Yeah, frankly I'm not a fan of this. That's wrapping a lot of operations in a single function. We have some loess stuff already, but specifying a whole regression model inside `tinyplot`?! I don't think `plt()` should become a general model-fitting interface.

And I'm not sure I understand what `.tinyplot` S3 method is supposed to do. What object class is `parameters` expecting to process? AFAIK, there isn't a `tinyplot` object we can pass to a `parameters_model` function.

vincentarelbundock:
Also, the natural workflow is:

1. Prep data
2. Fit model
3. Extract coefficients
4. Plot

It would be weird for `parameters` to implement something that is "downstream" from them.

grantmcdermott:
[...]

PR #571, opened 2026-04-23
type_mean_se

marcboschmatas:
Some time ago, I opened an issue to see whether it was possible to create a "mean_se" option for `type_summary()` (https://github.com/grantmcdermott/tinyplot/issues/436). You mentioned this was difficult, and I've been working on it for some time. The best solution I could come up with was to make a `type_mean_se() ` (https://github.com/marcboschmatas/tinyplot/blob/main/R/type_mean_se.R). I managed to do one which works well enough, but it has a pretty important problem here, instead of getting the categorical values for X, I get their factor order, so for instance,

`plt(bill_length_mm ~ species, data = penguins, type = "mean_se")`
returns
imatge

Trying to pass character values to draw_pointrange() will return this error.

```

Error in segments(x0 = ixmin, y0 = iymin, x1 = ixmax, y1 = iymax, col = icol,  : 
  first argument is not valid
Called from: segments(x0 = ixmin, y0 = iymin, x1 = ixmax, y1 = iymax, col = icol, 
    lwd = ilwd)
``` 

EDIT: this has been sorted now, as well as dodging issues, so this code

```r
plt(mpg ~ cyl | gear, data = mtcars, type = type_mean_se(dodge = .1))
```
Returns this plot
imatge

I don't know if you'd be interested in incorporating this feature. Of course, I'll incorporate formal tests and so on before merging.

Thanks!

DISCLAIMER, I have used Claude AI for some debugging.


grantmcdermott:
Thanks @marcboschmatas. It will be a little while before I can evaluate this in detail. But for the moment, I'll just say that what you're proposing here is not too dissimilar for a `"coefplot"` type that I suggested some time ago. https://github.com/grantmcdermott/tinyplot/issues/382

I dropped pursuing it b/c of lukewarm feedback (@vincentarelbundock and @zeileis both had good pushback). Maybe the situation has evolved since then, but I'll let others comment in the interim.

vincentarelbundock:
Yeah, this is for sure a useful function. However, I'm still not convinced it's a good idea for `tinyplot`, for the same reasons at the link above. Sorry to not be more positive!

marcboschmatas:
Hi @grantmcdermott and @vincentarelbundock , I've read through #382 and I can see why it's a bit more complicated than I thought, and I mostly agree with Vincent's line of thought. If you see this going somewhere, we can leave it open, otherwise, I'm OK with closing it, and I'll look for something to help with anything else in tinyplot that needs doing, it's a great package, and I'd realy like to contribute a bit.

grantmcdermott:
Thanks @marcboschmatas. Yes, I think it's better to close then, esp. given the spate of recent changes to the codebase which would be a pain to merge. Sorry for the 'wasted' effort, but hopefully you got to familiarize yourself with the `tinyplot` codebase a bit more. Speaking of which...

Yes, we'd welcome future contributions! I'd recommend holding off a bit until `0.7.0` is released, since I'm bundling together a bunch of aesthetic improvements. But feel free to look through the open issues after that and pick anything that takes your fancy.

P.S. I recently learned of the [forrest](https://github.com/lorenzoFabbri/forrest) package, which is built on top of tinyplot and might address part of the specific gap that you were looking to fill here. (Or, not... feel free to create your own "tinycoefplot" derivative ;-) )

Issue #409, opened 2025-05-20
Improve `tinytheme("ipsum")`?


grantmcdermott:
I was playing around with some code yesterday and realised that our "ipsum" theme implementation is actually quite far off the original. Compare:

``` r
library(ggplot2)
library(hrbrthemes)
library(tinyplot)

ggplot(mtcars, aes(mpg, wt, col = factor(cyl))) +
  geom_point() +
  labs(x="Fuel efficiency (mpg)", y="Weight (tons)", 
       title="Seminal ggplot2 scatterplot example",
       subtitle="A plot that is only useful for demonstration purposes"#,
       # caption="Brought to you by the letter 'g'"
       ) + 
  theme_ipsum() +
  scale_colour_ipsum()
```

![](https://i.imgur.com/ndeb7tc.png)

``` r
tinytheme("ipsum")
plt(
  wt ~ mpg | factor(cyl), mtcars,
  xlab="Fuel efficiency (mpg)", ylab="Weight (tons)", 
  main="Seminal ggplot2 scatterplot example",
  sub="A plot that is only useful for demonstration purposes"
)
```

![](https://i.imgur.com/asnvx7G.png)

While our theme implementations are more in the spirit of "inspired by" than intended to be exact replicas, bringing them closer into alignment only requires a few tweaks:

``` r
tinytheme("ipsum", family = "Arial Narrow", font.main = 2, font.sub = 3,
          palette.qualitative = ipsum_pal()(9))
plt(
  wt ~ mpg | factor(cyl), mtcars,
  xlab="Fuel efficiency (mpg)", ylab="Weight (tons)", 
  main="Seminal ggplot2 scatterplot example",
  sub="A plot that is only useful for demonstration purposes"
)
```

![](https://i.imgur.com/kPy2c5m.png)

Of the above changes, the following would be very straightforward to implement:
- Change main and sub font families. (Optional, but also easy: Reduce size of axis titles and tick labels)
- Use the same colors for `palette.qualitative` as `ipsum_pal()(9)`

Changing the default family to "Arial Narrow" is more complicated, since I'm not fully up to speed on precisely how wide the expected coverage is. (Maybe @hrbrmstr or @eddelbuettel could weigh in based on their experience?) But we could perhaps implement some internal logic that checks for an installed version of Arial Narrow first, before defaulting back to "sans".

[...]

PR #578, opened 2026-05-08
Support `grid = "x/y/X/Y"`

### LINKED ISSUE #193: grid = grid() not working properly with faceted plots
Specifically, it only seems to be working for the first facet. Tweaked example from the README, where we use `grid = grid()` instead of `grid = TRUE`.

``` r
library(tinyplot)
plt(
  Sepal.Length ~ Petal.Length | Sepal.Length, data = iris,
  facet = ~Species, facet.args = list(bg = "grey90"),
  pch = 19,
  main = "Faceted Species!",
  grid = grid(), # <--
  frame = FALSE
)
```

<sup>Created on 2024-08-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### LINKED ISSUE #577: Finer panel `grid` lines
Originally raised by @zeileis here: https://github.com/grantmcdermott/tinyplot/issues/409#issuecomment-2896126598

> Also, can we easily set a strategy to obtain the finer grid (with one extra `grid` step in between the ticks)? I actually do like that better than the rather coarse grid that we get in tinyplot.

We need to think about both the 1) user interface and 2) internal logic.

1. For the user-facing interface, I quite like the `grid = "xy"` vs `grid = "XY"` input logic that Bob Rudis uses in his [hrbrthemes](https://github.com/hrbrmstr/hrbrthemes) package. The only difference is that I wouldn't worry about combination variants, e.g., `"xXY"`. I would just stick to `<lower case> => fine grid`, while `<upper case> => regular grid` (where the latter is equivalent to our current `grid = TRUE` default). Similarly, I wouldn't worry about different tickness for the minor grid elements. Just keep everything the same.

2. For internal implementation, we would need to edit the [panel grid logic](https://github.com/grantmcdermott/tinyplot/blob/49e925980fd0bdc50d5d9e362f17e7733f070ed9/R/facet.R#L515) of `facet.R`, so that it captures these finer grid cases and calculates the appropriate (halved) break points. Something like this as a manual example:

``` r
library(tinyplot)
tinytheme("clean2", grid = FALSE)

plt(wt ~ mpg | factor(cyl), mtcars)

xaxbs = 5/2; xlim = par('usr')[1:2]
yaxbs = 1/2; ylim = par('usr')[3:4]

abline(
  v = seq(floor(xlim[1] / xaxbs) * xaxbs, ceiling(xlim[2] / xaxbs) * xaxbs, xaxbs),
  col = tpar('grid.col'), lwd = tpar('grid.lwd'), lty = tpar('lty')
)
abline(
  h = seq(floor(ylim[1] / yaxbs) * yaxbs, ceiling(ylim[2] / yaxbs) * yaxbs, yaxbs),
  col = tpar('grid.col'), lwd = tpar('grid.lwd'), lty = tpar('lty')
)
```

<sup>Created on 2026-05-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

grantmcdermott:
Closes #577 
Closes #193 

Also makes it easy to add only vertical or horizontal grid lines, which opens up support for some new themes... e.g. broadsheet publications like _Financial Times_, _Economist_, etc. tend to use single axis grid lines in all of their plots.

Some examples:

``` r
#> ℹ Loading tinyplot

plt(wt ~ mpg | factor(cyl), mtcars, grid = 'xy')
```

``` r
plt(wt ~ mpg | factor(cyl), mtcars, grid = 'xY')
```

``` r
plt(wt ~ mpg | factor(cyl), mtcars, grid = 'Y')
```

And, of course, the main attraction is using a finer grid for certain themes. Here's a riff on our proposed improvements to the `"ipsum"` theme in #409.

``` r
tinytheme(
  "ipsum", family = "Arial Narrow", font.main = 2, font.sub = 1, cex.lab = 0.8,
  palette.qualitative = c("#D18975", "#8FD175", "#3F2D54", "#75B8D1", "#2D543D", "#C9D175", "#D1AB75", "#D175B8", "#758BD1"),
  tcl = -0.0,  mgp = c(1.5, 0.5, 0) - c(0.0+0.1, 0.1, 0),
  grid = "xy"
)

plt(
  wt ~ mpg | factor(cyl), mtcars,
  xlab="Fuel efficiency (mpg)", ylab="Weight (tons)", 
  main="Seminal tinyplot scatterplot example",
  sub="A plot that is only useful for demonstration purposes"
)
```

``` r
tinytheme()
```

<sup>Created on 2026-05-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


grantmcdermott:
P.S. I can't add test snapshots right now, since my devcontainer setup is acting up for some reason.

grantmcdermott:
Okay, I think I've managed to catch all remaining edge cases and simultaneously fix a minor, but long-standing bug (#193).

PR #613, opened 2026-06-07
added tinyplot.data.frame method

zeileis:
Replaces #347 

The new `data.frame` method for `tinyplot()` draws some inspiration from the corresponding method for base `plot()`. It mainly facilitates using `tinyplot()` with data frames and pipes.

```
tinytheme("clean2")
cars |> tinyplot()
```

tinyplot-dataframe1

```
iris |> tinyplot(Sepal.Length ~ Petal.Width | Species)
```

tinyplot-dataframe2
     
Currently, this only works for data frames with 1 or 2 variables. In the future we might add a pairs-style display as follows but this would require better handling of axes and their labels.

``` 
par(mfrow = c(5, 5))
for (i in names(iris)) for(j in names(iris)) tinyplot(iris[, unique(c(j, i)), drop = FALSE])
```

tinyplot-dataframe3

So far the PR only contains code and documentation. NEWS and tests can be added when we agree on the code.


grantmcdermott:
This is great, thanks @zeileis.

As you would have seen, your updates here have inspired me to tack a crack at fixing the default colour case in #614. I think we should agree on that PR first and, if we decide to adopt it, merge those changes into this branch so we can review the combined effect.

R/tinyplot.data.frame.R
```diff
@@ -0,0 +1,58 @@
...
+#'   respectively. For data frames with more than 2 variables the `facet = NULL`
+#'   case is not supported, yet, and currently leads to an error.
+#' @param ... further arguments passed to `tinyplot`. 
+#'
+#' @examples
+#' tinytheme("clean2")
+#' 
+#' ## using tinyplot() with data frames and pipes
+#' cars |> tinyplot()
+#' iris |> tinyplot(Sepal.Length ~ Petal.Width | Species)
```
grantmcdermott: Using the `|>` in our examples is going to trigger a requirement upgrade to R>=4.1.0 on CRAN's system. I'd like to preserve our current R>=4.0.0 requirement. So, I suggest we either comment these examples out (with an appropriate remark explaining to users), or wrap everyting in `\dontrun{}`.

R/tinyplot.data.frame.R
```diff
@@ -0,0 +1,58 @@
...
+tinyplot.data.frame = function (x, formula = NULL, ...) {
+  ## original call
+  cl = match.call()
+
+  ## default formula
+  ## 1d: y ~ 1
+  ## 2d: y ~ x
+  ## 3d: pairs? (not supported, yet)
+  if (is.null(formula)) {
+    if (ncol(x) > 2L) stop("'formula' is missing with no default for more than 2 columns")
```
grantmcdermott: Hmmm, I think we could do better than erroring out here. My own preference is that we either:

1) Subset to the first 2 columns and display the resulting simple two-way plot (maybe with a message), or

2) Subset to the first N (<=5?) columns and then internally implement the "manual" pairs-esque plot that you demonstrate in the Examples above (maybe with a message if any columns are dropped), or

3) Don't subset at all and just try to implement the "manual" pairs-esque plot you demonstrate above (but risk poor viewability/zooming.)

P.S. I don't think we should try to mimic base `plot(df)`/`pairs(df)` exactly. Apart from the top + right axes variation being a pain, I think your example shows several improvements, especially w.r.t. to the histogram and way that factor/character variables are handled consistently (barplots + spineplots).

P.P.S. If we go with option 2 or 3, one other thing to consider is whether we support an (atomic) `by` argument without a formula. E.g., `tinyplot(iris, by = "species")`.

R/tinyplot.data.frame.R
```diff
@@ -0,0 +1,58 @@
...
+#' cars |> tinyplot()
+#' iris |> tinyplot(Sepal.Length ~ Petal.Width | Species)
+#'
+#' ## tinyplot(df) only works for data frames with 1 or 2 variables
+#' ## in the future we might add a pairs-style display as follows
+#' ## but this would require better handling of axes and their labels
+#' par(mfrow = c(5, 5))
+#' for (i in names(iris)) for(j in names(iris)) tinyplot(iris[, unique(c(j, i)), drop = FALSE])
+#'
+#' tinytheme() ## reset
```
grantmcdermott: We also need to reset `mfrow` here, e.g. (see above): `par(op)`

R/tinyplot.data.frame.R
```diff
@@ -0,0 +1,58 @@
...
+#' 
+#' ## using tinyplot() with data frames and pipes
+#' cars |> tinyplot()
+#' iris |> tinyplot(Sepal.Length ~ Petal.Width | Species)
+#'
+#' ## tinyplot(df) only works for data frames with 1 or 2 variables
+#' ## in the future we might add a pairs-style display as follows
+#' ## but this would require better handling of axes and their labels
+#' par(mfrow = c(5, 5))
+#' for (i in names(iris)) for(j in names(iris)) tinyplot(iris[, unique(c(j, i)), drop = FALSE])
```
grantmcdermott: I like this example, but would tweak it as:

```r
nms = names(iris)
op = par(mfrow = rep(length(nms), 2))
tinytheme("clean2", cex = 0.67)
for (i in nms) for(j in nms) {
    tinyplot(iris[, unique(c(j, i)), drop = FALSE])
    box("figure", lwd = 2)
}
```

R/tinyplot.data.frame.R
```diff
@@ -0,0 +1,58 @@
+#' tinyplot Method for Plotting Data Frames
+#'
+#' @description Convenience interface for visualizing data.frame objects
+#'   with tinyplot.
+#'
+#' @details This is a convenience function for plotting data frames with
+#'   or without a formula. The case with the formula mainly facilitates
+#'   using `tinyplot()` in combination with pipes. The case without
+#'   formula provides a quick way of plotting the variables in data frames,
+#'   either only one variable or a pair of variables. In the future, the
```
grantmcdermott: See my comment about the n>2 case below.

grantmcdermott (commented):
Some specific (and high-level) thoughts in my review below. Happy to discuss!

grantmcdermott:
@mthulin please take a look too, since this would replace your previous PR in #347.

R/tinyplot.data.frame.R
zeileis: Good point, thanks! Wrapping it in `if (getRversion() >= "4.1.0")` should also work, I think?

R/tinyplot.data.frame.R
zeileis: You are right. I think we should just go for 3 and explain the limitations in the manual page.

R/tinyplot.data.frame.R
zeileis: The `getRversion` check is not enough because the parser still detects the `|>` operator. I've commented the examples now.

R/tinyplot.data.frame.R
zeileis: I have removed the example now because the pairs-esque display is now built into `tinyplot.data.frame`.

R/tinyplot.data.frame.R
zeileis: I have removed the example now because the pairs-esque display is now built into `tinyplot.data.frame`.

R/tinyplot.data.frame.R
zeileis: I have attempted to implement 3 now including some very simple setting of `xlab`, `ylab`, and `main`. It's not great but probably better than nothing. Further improvements would be welcome.

R/tinyplot.data.frame.R
zeileis: I have updated it a little bit now. But depending on where we go with the pairs-esque display, further updates will probably be necessary.

grantmcdermott:
Hi @zeileis,

I made a few tweaks for the n > 3 case, which are more suggestions that than definitive choices. But want to push so you could test on your side. Let me know what you think (and feel free to push back if you feel strongly about anything):

- I dropped the axis labs for all of the sub plots. This is consistent with how `pairs` does it and also means that the plot regions are all equally size.
- I scaled the `cex` for the individual plots, so that the points etc. are a bit smaller. This is similar to how we reduce the `cex` size in faceted plots (and base R plots are also scaled for `mfrow` adjustments).
- I added a box around each sub-plot. I just found that this helped me a bit visually; especially for certain themes. But maybe we can add an argument so that users can control this.

MWE:

```r
plt(iris, theme = 'clean2')
```

df-iris

zeileis:
An important (and deliberate) difference between our plot and `pairs` is that we have a univariate display on the diagonal while `pairs` only has a large label. That's why `pairs` doesn't need additional axis labels. In our plot I felt that the axis labels in the margins were a useful addition because the main labels on the diagonal are a bit smaller. Maybe we should control this by an extra argument?

The `cex` looks good to me. Good idea.

Personally, I'm not fond of the frames. But maybe an argument would also be useful here.

Finally, you might want to have a stab at the additional `by` argument? This would also be very useful.

grantmcdermott:
Done. I added `labs` and `frames` args to control the axis titles and subplot boxes, respectively. I also added support for a `by` argument.

I need to get to bed but please kick the tyres and let me know if this looks good to you now.

P.S. On the axis labels/titles. I initially tried to support an "outer" subplot labeling only; what you were referring to "margin" labels above. The trouble is that there's no way to do this consistently, In the `iris` dataset, for example the top left plot y-axis ends up saying "Frequency", while the rest of the top row y-axis are really representing "Sepal.Width". So I went for a simple all-or-nothing approach.

zeileis:
Just had a very quick look, very nice!! Two short comments:

- In themes that have axis lines (such as the default theme), I think that `frames = TRUE` is definitely too much. For other themes, frames are ok, even if I typically prefer no frames. Hence the simplest solution would be to set the default to `frames = FALSE`.

- In addition to `by` being a character string indicating an existing column, we could also allow a full variable so that you can do: `tinyplot(iris[, 1:4], by = iris$Species)`.

grantmcdermott:
> Two short comments:

Done! I think that this is ready to be merged now.

zeileis:
Thanks, this is fantastic.

One more thought/question: Should we separate the functionality into a `tinypairs()` function as well? This would facilitate the creation of 2 x 2 pairs displays.

grantmcdermott:
> One more thought/question: Should we separate the functionality into a `tinypairs()` function as well? This would facilitate the creation of 2 x 2 pairs displays.

Sure, that sounds reasonable.

PR #639, opened 2026-06-21
Top-level (NSE) support for `weights` and `labels`

### LINKED ISSUE #332: Support weights

grantmcdermott:
Closes #332

Adds two new, top-level `tinyplot()` arguments: 

- `weights` arg only passes through for the model-fit types (e.g., `"lm"`, `"glm"`) and distribution types (e.g., `"histogram"`, `"spineplot"`). Passing through a top-level `weights` arg for a non-supported type is ignored, but triggers a warning (since it implies an unused statistical transformation).
- `labels` arg only passes through for `type_text` and is silently ignored otherwise.

Both of these are supported via NSE from the top-level, but users can also supply an equivalent atomic vector argument directly through the relevant `type_*()` constructor. For example, these two lines are equivalent:

```r
plt(y ~ x, dat, type = "lm", weights = w)       # top-level, nse
plt(y ~ x, dat, type = type_lm(weights = dat$w) # type-level, vector
```

(However, the bottom variant is safer b/c it is incorporated into the `model.frame` computation, and so will correctly account for `NA`s across input columns, as well as `subset` operations.)

If both forms are provided at the same time, then the top-level variant wins out.

## MWE

``` r
#> ℹ Loading tinyplot
s77 = as.data.frame(state.x77)
plt(`Life Exp` ~ Income, data = s77)
plt_add(type = "lm")
plt_add(type = "lm", weights = Population, col = "red")
```

``` r
ttnc = as.data.frame(Titanic)
plt(
    Survived ~ Sex,
    data = ttnc,
    type = "spineplot",
    weights = Freq
)
```

``` r
aq = transform(airquality, mnth = month.abb[Month])
# plt(Ozone ~ Temp, data = aq, type = type_text(labels = aq$mnth)) ## would fail b/c of NAs in Ozone
plt(Ozone ~ Temp, data = aq, type = "text", labels = mnth)
```

<sup>Created on 2026-06-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


grantmcdermott:
@zeileis @vincentarelbundock Not to nag or rush, but please let know if you have any conceptual objection to me merging this (or would like to test / kick the tyres before giving the go-ahead).

zeileis:
I cannot keep up with your pace, Grant! 😳

But I gave this a quick trial now and noticed nothing problematic. In combination with other recent PRs we can now easily generate doubledecker plots from tabular input:

```
UCBAdmissions |>
  as.data.frame() |>
  tinyplot(Admit ~ Gender, facet = 1 ~ Dept, weights = Freq)
```

tinyplot-doubledecker

grantmcdermott:
> In combination with other recent PRs we can now easily generate doubledecker plots from tabular input:

Oh, that's a cool example. I'll add it to the documentation.

PR #641, opened 2026-06-22
Fix no annotation issue 

### LINKED ISSUE #626: ann=FALSE is ignored
From `?plt`:

> ann: a logical value indicating whether the default annotation (title and x and y axis labels) should appear on the plot.

```R
plot(0,0,ann=FALSE)  # no axis labels
 plt(0,0,ann=FALSE)  # does have axis labels
```

zeileis:
Fixes #626 

For compatibility with base R `tinyplot(..., ann = FALSE)` should suppress title and axis notation.

This is implemented now by passing the `ann` argument to `draw_title()` and only drawing `title()` and `mtext()` elements if `ann = TRUE`.

### DIFF
--- a/R/tinyplot.R
+++ b/R/tinyplot.R
@@ -1244,7 +1244,8 @@ tinyplot.default = function(
       }
     }
 
-    draw_title(main, sub, cap, xlab, ylab, legend, legend_args, opar,
+    ann = as.logical(ann)
+    draw_title(main, sub, cap, xlab, ylab, legend, legend_args, opar, ann,
                xlab_line_offset = if (!is.null(dynmar_computed)) .whtsbp_x_raw else 0,
                ylab_line_offset = if (!is.null(dynmar_computed)) .whtsbp_y_raw - max(0, .ymgp_shift) - .ylab_cex_shift else 0)
   }
--- a/R/tinyplot.R
+++ b/R/tinyplot.R
@@ -1245,9 +1245,9 @@ tinyplot.default = function(
     }
 
     ann = as.logical(ann)
-    draw_title(main, sub, cap, xlab, ylab, legend, legend_args, opar, ann,
-               xlab_line_offset = if (!is.null(dynmar_computed)) .whtsbp_x_raw else 0,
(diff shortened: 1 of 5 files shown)


R/tinyplot.R
```diff
@@ -1244,7 +1244,8 @@ tinyplot.default = function(
       }
     }
 
-    draw_title(main, sub, cap, xlab, ylab, legend, legend_args, opar,
+    ann = as.logical(ann)
+    draw_title(main, sub, cap, xlab, ylab, legend, legend_args, opar, ann,
```
grantmcdermott: Could we just use `if (ann) draw_title(...)` here and skip the individual conditions in `draw_title.R` below?

R/tinyplot.R
```diff
@@ -1244,7 +1244,8 @@ tinyplot.default = function(
       }
     }
 
-    draw_title(main, sub, cap, xlab, ylab, legend, legend_args, opar,
+    ann = as.logical(ann)
```
grantmcdermott, line 1247: Nit: I think we should use `assert_logical(ann)` (no need to re-assign).

grantmcdermott (changes requested):
Thanks @zeileis. Two comments, one small and the other a potential efficiency gain. LMK!

R/tinyplot.R
zeileis: Probably, yes. I wasn't sure whether any of the computations in `draw_title()` had any side effects that I might have overlooked.

But going through the code once more, I think you're right. Will try and run `R CMD check`.

R/tinyplot.R
zeileis: In base R `ann = 1` works like `ann = TRUE`. Hence, I thought that we should do the same.

grantmcdermott: Fair enough. Stick with what you have then, thanks.

R/tinyplot.R
zeileis: I couldn't find any negative side effects for the simpler solution, pushed now.

grantmcdermott:
Thanks!

PR #657, opened 2026-07-02
Consistent fill lighten for area types

### LINKED ISSUE #646: Area-based plot types: light opaque multi-group color
In the discussion of #614 we decided to use lighter opaque (non-transparent) colors for area-based displays. However, it seems like we only implemented this for the single-group case but not the multi-group case.

In the multi-group case we still get semi-transparent (and hence light) colors for histograms and boxplots and dark opaque colors for barplots. Hence, the difference is really striking for barplots:

```
library("tinyplot")
tinytheme("clean2")
tinyplot(~ species, data = penguins)
tinyplot(~ species | island, data = penguins)
```

Image

For boxplots the difference is more subtle but if you know it it's hard to overlook it in themes with grid lines:

```
tinyplot(body_mass ~ species, data = penguins)
tinyplot(body_mass ~ species | sex, data = penguins)
```

Image

For the multi-group histograms one could argue, though, that semi-transparency is the right thing because the distributions will often be overlapping:

```
tinyplot(body_mass ~ 1, data = penguins)
tinyplot(body_mass ~ 1 | sex, data = penguins)
```

Image

grantmcdermott:
Closes #646

Using the same summary that I shared in my [earlier comment](https://github.com/grantmcdermott/tinyplot/issues/646#issuecomment-4861074172):

* Grouped `barplot`, `boxplot`, and `violin` now use the same lighter-but-opaque fill as their single-group counterparts. This fixes the the original inconsistency that you pointed out at the top of this issue (dark saturated bars vs. semi-transparent boxes).
     * The way that this works is that each of these types gains a `lighten` argument (default TRUE). So, for example, you could override with `type_barplot(lighten = FALSE)` to fall back to the old saturated palette colour.
     * `spineplot` also gets a similar lightening treatment, but only for the special `y == by` case. I've also made sure that these get consistent (black) borders regardless of grouping. 
* The alpha-darkening bug is fixed. We now resolve the palette opaque, lighten, and then layer any transparency on top.s fixed. 

**MWE**

Using some of the examples from the original issues:

``` r
#> ℹ Loading tinyplot
tinytheme("clean2")
tinyplot(~ species, data = penguins)
```

``` r
tinyplot(~ species | island, data = penguins)
```

``` r
tinyplot(body_mass ~ species, data = penguins)
```

``` r
tinyplot(body_mass ~ species | sex, data = penguins)
```

``` r
tinyplot(~ species, data = penguins)
```

``` r
tinyplot(~ species, data = penguins, fill = 0.7)
```

<sup>Created on 2026-07-01 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


zeileis:
To support specifications like `lighten = 0.5` in addition to just `lighten = TRUE` we could use:

```
lighten_col = function(x, amount = 0.8) {
  x = to_hcl(x)
  hcl(
    h = x[, "H"],
    c = x[, "C"] * (1 - amount),
    l = 100 - (100 - x[, "L"]) * (1 - amount),
    alpha = attr(x, "alpha")
  )
}
```

By default this yields similar colors as the current `lighten_fill()` which is based on `seq_palette(..., n = 3)[3]`. Because we just need the lightened color here (rather than a palette going from the full color to the lightened one), `lighten_col()` can be a little bit simpler and hence leads to slightly different output.

Additionally, we may want to build in a catch that makes sure that `lighten_col("black")` returns `"lightgray"` to be exactly consistent with base R.

grantmcdermott:
Thanks @zeileis, I think that both `lighten = [0,1]` and `lighten_col` are good ideas... But I'm afraid that I'm up against a hard deadline and so will have to defer to until a later date.

(_tl;dr_ I'm leaving for 3 weeks vacation, with limited computer access. But I _have_ to get this next CRAN submission out first, so that some work colleagues can easily use v0.7.0 for an internal project while I'm gone. It's a long story, but they need some our latest features and also have to install from the official CRAN distribution for some infosec reasons.)

Anyway, the bottom line is that I need to merge this today if it's going to make the v0.7.0 submission. IMHO it should be included if at all possible and I also feel it's basically in good enough shape now. But I can hold off until later today if you need/want to confirm anything else on your side.

Let me know!

grantmcdermott:
P.S. I realize that I didn't include any spineplot **MWEs** above. I was probably unnecessarily confusing in my previous comment https://github.com/grantmcdermott/tinyplot/issues/646#issuecomment-4861074172... The takehome version is that, for most cases, spineplots are essentially unchanged. The only user-facing changes that I implemented were:

1. We now draw black borders by default for all spineplots; whereas we previously turned these off automagically for some plots. Note that users can still turn borders off manually, by calling (say) `lty = 0`. This feels more principled to me.
``` r
#> ℹ Loading tinyplot
tinytheme("clean2")
tinyplot(
  Species ~ Sepal.Length, data = iris, 
  type = "spineplot"
)
```



``` r
tinyplot(
  Species ~ Sepal.Length, data = iris,
  type = "spineplot", lty = 0
)
```



2. Users have the option to call `lighten = TRUE` for spineplots where `y==by` (and only for this narrow case). This argument is set to `FALSE` by default, but I felt it made sense to add the option give our other updates to `barplot` and co. above (although the default for these other types is obviously `TRUE`).

``` r
tinyplot(
  Species ~ Sepal.Length | Species, data = iris,
  type = "spineplot"
)
```



``` r
tinyplot(
  Species ~ Sepal.Length | Species, data = iris,
  type = type_spineplot(lighten = TRUE)
)
```

Issue #299, opened 2025-02-03
Easy way to reverse palette?


grantmcdermott:
I might be missing a workaround, but I've encountered a few situations where I'd like an easy way to reverse the colour palette. For example, here the colours of the "temps" palette are running in the wrong direction 

``` r
library(tinyplot)

plt(Month ~ Temp, airquality, type = type_ridge(gradient = "temps"))
```

![](https://i.imgur.com/xkBprqr.png)

<sup>Created on 2025-02-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

grantmcdermott:
One workaround is manually specifying the palette and passing the `rev = TRUE` argument. But it isn't particularly convenient.
``` r
library(tinyplot)
plt(Month ~ Temp, airquality, type = type_ridge(gradient = TRUE),
    palette = hcl.colors(n = 100, palette = "temps", rev = TRUE))
```

![](https://i.imgur.com/Bew1IWv.png)

<sup>Created on 2025-02-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

zeileis:
I couldn't find it in the issues but I vaguely recall that we had a similar discussion about a convenience option for dropping the first color in a palette (which is often useful if the first color is black). But we didn't go for it because it was already possible by passing a vector of colors to `palette`.

I'm not completely sure, though, whether I remember this correctly...

PR #677, opened 2026-08-02
`type = "tile"` and `type = "heatmap"`

### LINKED ISSUE #665: `"heatmap"`

grantmcdermott:
Closes #665

**Edit:** Modified to drop the `"heatmap"` alias. We now treat `type_heatmap()` as its own related, but distinct type.

Some quick examples from the documentation: 

``` r
# library(tinyplot)
#> ℹ Loading tinyplot

# It is recommended to use the dedicated "heatmap" theme for this type
tinytheme("heatmap")

# Correlation matrix of the base `attitude` dataset in "long" form.
catt = as.data.frame(as.table(cor(attitude)), responseName = "Correlation")

tinyplot(Var1 ~ Var2 | Correlation, data = catt, type = "tile")
```

``` r

# slightly fancier version, where we suppress the legend but layer on the values
# as text
tinyplot(
  Var1 ~ Var2 | Correlation, data = catt,
  type = "tile",
  legend = FALSE,
  xlab = NA, ylab = NA,
  main = "Correlation matrix of base attitude dataset"
)
tinyplot_add(type = "text", labels = round(catt$Correlation, 2), col = "white")
```

``` r


``` r
# Pass scaled tile widths and heights through type_tile() for a gridded look
tinyplot(
  Var1 ~ Var2 | Correlation, data = catt,
  type = type_tile(width = 0.9, height = 0.9)
)
```

``` r
# default palette is tealgrnn, but easy to pass through a diverging scale if
# your dataset recommends it
m = as.data.frame(as.table(cor(mtcars)), responseName = "Correlation")
tinyplot(
  Var1 ~ Var2 | Correlation, data = m, type = "heatmap",
  palette = "tealrose", col = "white" # white rect borders (optional)
)
```

``` r

# restore theme
tinytheme()
```


zeileis:
Some very quick feedback only based on the examples above. (I didn't have time to look at the code or try the examples myself, yet.)

- I like the name `"tile"` because I feel `"heatmap"` is too ambiguous.
- Possibly a future `"bkde"` plot could just set up the data in the format needed and then plot it via "tile".
- What happens if the same tile position is specified multiple times in the data? Are they simply overplotted?
- It might be important to control the mapping of the tile values to the tile colors. For example when plotting correlations, the center color of the diverging palette must be exactly at zero.
- When plotting correlation heatmaps the y-axis would usually be sorted top-down rather than bottom-up in order to match the layout of the matrix. Should this be handled by the user when preparing the data for plotting or should there be some convenience option for it?

grantmcdermott:
Thanks for the quick feedback @zeileis.

> * I like the name `"tile"` because I feel `"heatmap"` is too ambiguous.

You mean you'd like to drop "heatmap" even as an alias?

> * Possibly a future `"bkde"` plot could just set up the data in the format needed and then plot it via "tile".

👍

> * What happens if the same tile position is specified multiple times in the data? Are they simply overplotted?

Yes, we don't do any checks or transformations. (I think this is probably the right approach.)

> * It might be important to control the mapping of the tile values to the tile colors. For example when plotting correlations, the center color of the diverging palette must be exactly at zero.

Agree, but I haven't really thought how we could do that. Unless I'm missing something, we'd need some custom logic on top of `colorRamp` and co.? Would welcome other ideas if this is something you've done before. 

> * When plotting correlation heatmaps the y-axis would usually be sorted top-down rather than bottom-up in order to match the layout of the matrix. Should this be handled by the user when preparing the data for plotting or should there be some convenience option for it?

Hmmm. The user can just pass `x/ylim = "rev"`. I'm loathe to enable this by default, though since it would be messy trying to ID when to switch it off (e.g. non-correlation based heatmaps). But we could easily add an example to the docs.

grantmcdermott:
> > * It might be important to control the mapping of the tile values to the tile colors. For example when plotting correlations, the center color of the diverging palette must be exactly at zero.
> 
> Agree, but I haven't really thought how we could do that. Unless, I'm missing something, we'd need some custom logic on top of `colorRamp` and co.? Would welcome other ideas if this is something you've done before.

Actually, an easy solution is just to identify the midpoint and then expand the limits symmetrically around that midpoint. BUT the trick is deciding how users should be able to pass through this midpoint to the palette (`by` limits) logic...

On reflection, I think that this issue should be tackled in a separate PR as part of additional palette customization entry point. Maybe a sister, top-level `palette.args` argument for capturing and splicing in additional palette functionality? (C.f., https://github.com/grantmcdermott/tinyplot/issues/299)

zeileis:
I've been thinking about this some more and I agree that this should probably go into a separate issue/PR.

Is there currently the possibility to control the range of the `by` variable in the legend? I couldn't find one.

This would be another option to get the midpoint exactly at 0, namely by setting the `by` range to `c(-1, 1)`.

This would be useful more generally, e.g., when I want to use exactly the same `by` color scale for separate plots. When I have a factor `by` variable this happens automatically but for a continuous `by` variable I don't know whether this is currently possible.

zeileis:
Regarding the name: Yes, I would drop the `"heatmat"` alias as well.

grantmcdermott:
**Update:** I've dropped the `"heatmap"` alias... and instead separated `type_heatmap()` into its own related, but distinct type. The key difference vs. "vanilla" `type_tile()` is that `type_heatmap()` offers additional arguments for scaling (normalising) along either the `x` or `y` variables. This brings it much closer in line with with what the base `heatmap()` function does.

``` r
# library(tinyplot)
#> ℹ Loading tinyplot

tinytheme("heatmap")

mt = as.data.frame(as.table(as.matrix(mtcars)))

# first, the unscaled version: only `disp` and `hp` are visible
tinyplot(
  Var1 ~ Var2 | Freq, data = mt,
  type = "heatmap",
  xlab = NA, ylab = NA
)
```



``` r
# and now scaled within each x variable (i.e., column). The default is to
# z-score, matching base R's `heatmap(scale = "column")`.
tinyplot(
  Var1 ~ Var2 | Freq, data = mt,
  type = type_heatmap(scale = "x"),
  xlab = NA, ylab = NA
)
```



As an aside, note that we can also go "directly" through the `tinyplot.matrix` method, which avoids (the user) having to reshape the dataset long:

``` r
m = as.matrix(mtcars)
tinyplot(m, type = type_heatmap(scale = "x"), col = "white")
```



``` r

tinytheme()
```

<sup>Created on 2026-08-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

grantmcdermott:
P.S. To underscore the equivalence between base `heatmap()` and this new `tinyplot::type_heatmap()` implementation:

``` r
#> ℹ Loading tinyplot
m = as.matrix(mtcars)
heatmap(
  m,
  Rowv = NA, Colv = NA,
  scale = "column",
  main = "stats::heatmap()"
)
```



``` r
tinyplot(
  m,
  type = type_heatmap(scale = "x"),
  ylim = "rev",
  theme = list('heatmap', cex.axis = 0.75, palette.sequential = "YlOrRd"),
  main = "tinyplot::type_heatmap()"
)
```



<sup>Created on 2026-08-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Open questions (esp. interested in your thoughts @zeileis):

1. Are you happy (happier) with this additional `type_heatmap()` implementation as a separate type, with extra customizations on top of `type_tile()`?
2. To get a closer match to the base `heatmap()` look above, I had to tweak a few things. I wouldn't want to do all of these for `type_tile()` (since we should take inputs as literal with no meaningful adjusts/transformations). But should we do any of the following for `type_heatmap()`:
  a) Reverse the y-axis by default (`ylim = "rev"`)?
  b) Adjust the color palette from "tlgrn" to "YlOrRd""? (My vote is no, but I want to solicit feedback.)
  c) Adjust the axis labels (`axis.cex`) so that all categories display? (TBH, I'm not sure how to do this ahead of time...)

grantmcdermott:
> 2\. To get a closer match to the base `heatmap()` look above, I had to tweak a few things. I wouldn't want to do all of these for `type_tile()` (since we should take inputs as literal with no meaningful adjusts/transformations). But should we do any of the following for `type_heatmap()`:
>     a) Reverse the y-axis by default (`ylim = "rev"`)?
>     b) Adjust the color palette from "tlgrn" to "YlOrRd""? (My vote is no, but I want to solicit feedback.)
>     c) Adjust the axis labels (`axis.cex`) so that all categories display? (TBH, I'm not sure how to do this ahead of time...)


Okay, I've actioned 2a and 2c (albeit only in the sense that users can manually override the individual axes tick labels with the new `cex.xaxs` and `cex.yaxs` params). 2b is ignored. 

I'm pretty happy with the overall functionality now, as well as the separation of `type_tile`/`type_heatmap`, so am going to merge.

PR #678, opened 2026-08-04
flip = TRUE: flip the drawn geometry of "h" and the step types

### LINKED ISSUE #675: Support `type = "h"` with `flip = TRUE`
At the moment some single-letter plot `type`s, notably `type = "h"`, do not cooperate with `flip = TRUE`. Try:

```
tinyplot(1:10, (1:10)^2, type = "h", flip = FALSE)
tinyplot(1:10, (1:10)^2, type = "h", flip = TRUE)
```

Image

With `flip = TRUE` I would have expected the plot on the right, but I got the plot in the middle. The reason is that only the axes are flipped and `type = "h"` still draws vertical bars rather than horizontal bars. To obtain the desired output I had to use:

```
tinyplot(xmin = 0, ymin = 1:10, xmax = (1:10)^2, ymax = 1:10, type = "segments")
```

---

A similar issue exists with `type = "s"` and `"S"`, respectively. However, it is somewhat easier to work around this because when `flip = TRUE` one can simply switch to `type = "S"` and `"s"`, respectively.

```
tinyplot(1:10, (1:10)^2, type = "s", flip = FALSE)
tinyplot(1:10, (1:10)^2, type = "s", flip = TRUE)
tinyplot(1:10, (1:10)^2, type = "S", flip = TRUE)
```

Image

(If we wanted to mimic this behavior, we could implement a `type = "H"` for vertical histogram lines.)

haomeng797-ship-it:
Fixes #675.

`flip_datapoints()` swaps the coordinates, but `lines()` draws some single-letter types in a fixed orientation regardless of the data: `type = "h"` always drops vertically, and the step types commit to which coordinate moves first. So the axes flipped while the marks kept their unflipped geometry, which is the middle panel in the issue.

Changes in `draw_lines()`, all behind `isTRUE(flip)` so nothing else is touched:

- `"h"` now draws explicit horizontal segments from the baseline (0, or the left edge under `log = "x"`, matching how the base engine anchors its vertical drops).
- `"s"` and `"S"` swap under flip: reflecting a step path across the diagonal turns a move-then-rise step into a rise-then-move one, so the flipped rendering of `"s"` is `lines(type = "S")` on the swapped data, and vice versa. Both were silently drawing mirrored steps before.
- `"l"`, `"b"`, `"o"`, `"c"` are flip-invariant and unchanged.

On @zeileis's question about a `type = "H"` spelling: this PR only makes `flip` behave, so the axes stay swapped (value axis horizontal). A standalone `"H"` that keeps x as x would be a separate, additive decision, and nothing here gets in its way. Happy to do that as a follow-up if you want it.

Tests: rather than snapshots, the new tests in `test-flip.R` read the SVG output and assert the geometry directly (flipped `"h"` emits one horizontal `<line>` per point, unflipped stays vertical, and the first drawn step segment of flipped `"s"`/`"S"` moves in the right direction). They run on any OS, so they also exercise the fix on macOS where the snapshot suite skips.


grantmcdermott:
This is great, thanks @haomeng797-ship-it.

I'll let @zeileis test it, since he was the OP.

RE: tests. Your inversion test is clever... But I really would prefer dedicated snapshots. We do ship a dedicated `.devcontainer` for running and creating these Linux-bases snapshots on MacOS. I use [finch](https://github.com/runfinch/finch) on my own Mac for this:

``` sh
brew install --cask finch
finch vm init
finch vm start
```

Then, just open up VS Code in the directory of your locally cloned version of this repo. It should prompt you  to "Reopen in container" (say yes). To run the tests, open up a terminal in VS Code (`cmd` + `j`) and run:

``` sh
make testall
```

You'll probably get one or two false positive failures. But any new snapshots you generate should be good to go. More details here: https://github.com/grantmcdermott/tinyplot/blob/main/CLAUDE.md#testing--ci

Again, not a train smash if you can't do this. But useful to know, especially if you plan to continue making (these very useful!) PRs.

haomeng797-ship-it:
Thanks, that's really useful to know! Snapshots added 🙂 Four of them: `flip_type_h`, `flip_type_step`, `flip_type_step_rev`, and a grouped one. (Avoided `flip_type_s`/`flip_type_S` since those two filenames collide on a case-insensitive macOS volume.)

Ended up on colima rather than finch, since the finch cask wants a sudo password and colima installs straight from `brew install colima docker`. Same `rocker/r2u:24.04` image, so the SVGs come out byte-identical to the existing ones. Full `run_test_dir()`: 573 tests, 2 failures, and they're exactly the two false positives your CLAUDE.md warns about (`xaxl_yaxl` and `palette_manual_continuous`).

I left the SVG-geometry assertions in as well, mostly because they still run on macOS where the snapshots skip. Happy to drop them if you'd rather keep the suite snapshot-only, just say the word!

grantmcdermott:
Nice!

> I left the SVG-geometry assertions in as well, mostly because they still run on macOS where the snapshots skip. Happy to drop them if you'd rather keep the suite snapshot-only, just say the word!

Yes, please remove these since they're effectively duplicates now. Thanks.

haomeng797-ship-it:
Done — removed! Just the four snapshots now 👍

zeileis:
This is really cool, thanks Miura @haomeng797-ship-it ! Nice catch with with the `par("xlog")` case.

The fixes in the code look good to me and are ready to merge. With the test cases, I let Grant @grantmcdermott make the final decision.

zeileis:
**Addendum:** Apologies for taking so long to respond. While playing around with this I found some issues (now reported in #679)  that turned out to be unrelated to this PR. But I needed to boil down the examples first before I knew what was going on.

**Just for the recors:** In case anyone wonders how the two issues are related, this was the code I tried.

```
data("cns", package = "faraway")
tinyplot(I(An/NoCNS) ~ Area | Work, data = cns, type = "h",
  ylim = c(0, 0.0045), facet = "by", legend = FALSE, theme = "clean2", flip = TRUE)
tinyplot_add(type = "p")
```

tinyplot-cns1

The horizontal `type = "h"` from this PR works completely fine. It is the inconsistency between `type_points()` and `type_lines()` in the data handling that causes the problem. So this is really a separate issue.

I can generate the plot that I wanted if I swap the order of `type = "p"` and `type = "h"` and make sure that the ordering of the `Area` observations matches the ordering of its factor levels.

```
tinyplot(I(An/NoCNS) ~ factor(Area, levels = unique(Area)) | Work, data = cns, type = "p",
  ylim = c(0, 0.0045), facet = "by", legend = FALSE, theme = "clean2", flip = TRUE, xlim = "rev")
tinyplot_add(type = "h")
```

tinyplot-cns2

haomeng797-ship-it:
Thanks Achim, and no apology needed. That's a thorough test. Good to hear the xlog case was worth adding; I only noticed it because the baseline has to move when the axis is logged.

